PR #102, opened 2014-12-14
Add hook for pac editor config

suchipi:
Add a hook that runs at the bottom of pace's config.lua so that custom
parts can setup their editor config (location in tree, icon, etc) at the appropriate time


suchipi:
@CapsAdmin some of the newer pull requests expect this hook to be in place.

CapsAdmin:
Can you rename it to pac_EditorPostConfig just so it's consistent with all the other hooks?

suchipi:
Please merge https://github.com/CapsAdmin/pac3/pull/148 instead

PR #135, opened 2015-04-17
Patch add flex part

suchipi:
My interactive rebase decided it wanted to keep the comments, for some reason.

### DIFF
--- a/lua/pac3/core/client/netmessages.lua
+++ b/lua/pac3/core/client/netmessages.lua
@@ -39,3 +39,11 @@ function pac.SetCollisionGroup(ent,group)
 	net.WriteInt(group,7)
 	net.SendToServer()
 end
+
+function pac.TouchFlexes(ent)
+	local index = ent:EntIndex()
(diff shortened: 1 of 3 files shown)


CapsAdmin:
Here you give the freedom to call ent:SetFlexWeight(1,0) to any entity to every client. Are you sure this is okay?

suchipi:
If I remember correctly (wrote this a couple months ago), flex #1 is useless/harmless on all standard ragdolls/playermodels. I'll have to double-check though. All I remember for sure is wishing I didn't have to do this to get things working but concluding that it was "okay enough".

I'm gonna sleep for now, though; I'll follow up on this and the other PRs tomorrow as time permits

suchipi:
Conflicts resolved, updated to new postconfig hook. Also now we do ent:SetFlexWeight(1,ent:GetFlexWeight(1)). (The only reason to do this is so that the server starts tracking that that entity has modified flexes so that clients can update it).

PR #137, opened 2015-04-17
Patch custom animation part stuff

suchipi:


CapsAdmin:
conflict!

suchipi:
Resolved

Issue #130, opened 2015-04-05
Exploit with net message pac.net.SetCollisionGroup.ClientNotify


LUModder:
You can send a net message with this message with the right data and pretty much change the collision of other entities.

I will not post an example as someone might use this maliciously before it's patched.

thegrb93:
@suchipi You can't trust netmessages from the client to be legit.

I'm going to remove this function from my server until it gets fixed. Thanks for the report.

CapsAdmin:
I can totally see why. I didn't review the big merge properly.

In general I really like adding server features for reasons like these.

suchipi:
@thegrb93 submitted a PR to Metastruct/pac3 a day or two ago that just checks if the entity being modified is the player who sent the netmessage. I'll make a PR to bring it to upstream here soon; I'm behind for doing so for a few other features anyway (facial flexes, etc). @CapsAdmin would you prefer a bunch of small PRs in their own topical branches or one big one with squashed commits?

CapsAdmin:
smaller ones

CapsAdmin:
I'm not sure if this feature should be in pac or not. The way I see it it's kind of like adding godmode in pac. It seems like something an admin mod should do.

suchipi:
Does loading of costumes with part options that no longer exist behave gracefully? If so then removing this feature will be simple; I just don't want people to be unable to load costumes created in the months between its addition and removal.

CapsAdmin:
Nothing will happen. The next time they save their outfits that specific property will be gone from the save.

suchipi:
Did we remove this from your branch?

CapsAdmin:
yes

PR #138, opened 2015-04-17
Add check to collision group exploit.

suchipi:


suchipi:
Resolves https://github.com/CapsAdmin/pac3/issues/130

CapsAdmin:
I don't think this belongs in pac. At best it should be in the editor code. The core version of pac shouldn't have these features since gamemodes with strict rules are supposed to use it and leave the editor out.

This will still allow you to change your own collision whenever you want

CapsAdmin:
12:27 - Python1320: that is just exploit fix
12:27 - Python1320: if you want it gone you commit after it one that removes it

PR #139, opened 2015-04-17
fix default pac part icon

suchipi:

### DIFF
--- a/lua/pac3/editor/client/config.lua
+++ b/lua/pac3/editor/client/config.lua
@@ -661,5 +661,5 @@ function pace.TranslatePropertiesKey(key, obj)
 end
 
 function pace.GetIconFromClassName(class_name)
-	return pace.PartIcons[class_name] or "gui/silkicons/plugin"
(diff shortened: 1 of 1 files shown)


CapsAdmin:
Both icons exist and from what I understand the gui/* silkicons are deprecated in favor of icon16/*

What's the purpose of doing this?

suchipi:
I don't think the old path works at all anymore; before I made this change, parts that weren't in the tree would just display with a blank icon.

The only place I ever end up seeing it though is when developing parts in-game.

suchipi:
conflict (!?) resolved

suchipi:
Would you merge this if I bothered to rebase it?

CapsAdmin:
Yes. For some reason I read the merge the other way around (you changing it from "icon16/plugin.png" to "gui/icons/plugin)

suchipi:
I'm gonna just make a new branch and PR off of suchipi/pac3 because I don't know what's up with Metastruct/pac3 anymore

PR #141, opened 2015-04-17
Patch model part changes

suchipi:


suchipi:
@notcake @Python1320 please review this to make sure conflict resolution was handled properly

suchipi:
conflicts re-resolved

PR #145, opened 2015-04-17
silence error

suchipi:
false positive a lot of the time and spammy. You guys should really use
the webaudio part instead of the ogg part anyway

Conflicts:
	lua/pac3/core/client/parts/ogg.lua


suchipi:
This one was a community request...

CapsAdmin:
"if pac.debug then" maybe

suchipi:
That should work? (Unless pac.dprint accepts colors?)

PR #146, opened 2015-04-17
Testing PlayerSlowThink

suchipi:


suchipi:
@Techbot121 and @Python1320 might want to provide input on if this should even be open for a PR. https://github.com/Metastruct/pac3/issues/15 may also be relevant.

CapsAdmin:
python told me this requires PlayerSlowThink which I assume is metastruct specific. So unless you add PlayerSlowThink to pac this would be pointless to merge.

PR #149, opened 2015-04-19
Removed Dump SubMaterials

LUModder:
SetSubMaterial support is gone from this repo, so why even keep the tool in the editor?


CapsAdmin:
well it could maybe be used to see what materials a model has but whatever

the only use SetSubMaterial would have is in the model part. I could dynamically expand the material fields if the model supports more than one material.

PR #157, opened 2015-05-06
Fixed health_lost, added a few more events

edunad:
Added randint, allows you to specify on random a min and a max
Added random_timer event - A timer that triggers between a min and a max for x time.

fe8fea56-e9dd-11e4-964a-113928c719c1

ballsssssssssssssssssssssss

Fixed health_lost , randomly worked sometimes, the event was getting triggered too fast, din't let parents trigger. Had to add a manual cooldown, works now.

Added owner_alive - Checks if the owner is alive or not

### DIFF
--- a/lua/pac3/core/client/parts/event.lua
+++ b/lua/pac3/core/client/parts/event.lua
@@ -46,7 +46,50 @@ PART.Events =
 			return self:NumberOperator(math.random(), compare)
 		end,
 	},
+	randint = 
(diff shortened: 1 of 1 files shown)


Python1320:
If this hack does not work I'll revert and fix the old hack way, which is using timers. The problem right now was that Garry's timers were changed so that timer.Simple(0 runs on the same frame.

PR #173, opened 2015-06-03
Fix OGG playback for poorly manufactured headsets

TheFreezebug:
Adds a CVAR to fix OGG playback for the unfortunate people who bought headsets like the G930 or people who simply have incompatible soundcards.

The problem was never playing oggs through BASS, it was streaming them over URL.

The CVAR will cause PAC to download the entire file and play it from the disk (which is why this is optional and isabled by default)

### DIFF
--- a/lua/pac3/core/client/init.lua
+++ b/lua/pac3/core/client/init.lua
@@ -19,7 +19,7 @@ include("libraries/urlobj/urlobj.lua")
 include("libraries/urlobj/queueitem.lua")
 
 -- WebAudio
-include("libraries/webaudio/oggfix.lua")
+include("libraries/webaudio/ogg_fix.lua")
 include("libraries/webaudio/urlogg.lua")
 include("libraries/webaudio/browser.lua")
 include("libraries/webaudio/stream.lua")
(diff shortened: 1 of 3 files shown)


CapsAdmin:
the issue has always been "urlogg" and not bass afaik. i was hoping switching over to bass would "fix" it but i didnt know the bass implementation was so limited so forcing the ogg part to use bass was not feasible.

TheFreezebug:
I see what you mean with BASS, the implementation is very limited. I didn't look at URLOgg, i thought it worked directly through webaudio (sound.PlayURL) . Sorry for the bother!

CapsAdmin:
so bass doesn't work with headsets either then? i always assumed it was the
web audio processor stuff since pac does its own sound mixing in js with
the ogg part. i think i tried porting the relevant code to the web and got
people to test but that seemed to work fine. maybe it has to do with
resetting the soundcard? do these usb headsets disconnect when they go out
of range?

TheFreezebug:
@CapsAdmin Yeah, these USB headsets do go out of range when they disconnect. However, I think it has to do with the supported bitrate of the drivers for the headsets. Believe it or not, most of these headsets only support 16-bit mode audio. Another thing i've noticed is when you have a feature on USB headsets called "room fill" turned on, you are able to hear oggs from webaudio, but they sound absolutely awufl.

PR #198, opened 2015-07-19
Allow skins to do stuff with editor part node text colors.

Tenrys:


CapsAdmin:
For a skin that sounds oddly specific. I agree that the skin should decide the colors but this is what the line you commented out does. There are other areas of the code that need to change in order for that to happen.

Tenrys:
The thing is, that function changes how the color from a DTreeNodeButton is decided, where that isn't needed at all when you can just use skin colors.

--- Original Message ---

Python1320:
There should be Panel:SetDark(), can we use that instead? Feel free to reopen.

PR #213, opened 2015-09-18
Fixed the bodygroup part

TW1STaL1CKY:
Tested on model parts, players, npcs and other entities.
Also tested while in Singleplayer, (Local) Multiplayer, and in Metastruct.

### DIFF
--- a/lua/pac3/core/client/parts/bodygroup.lua
+++ b/lua/pac3/core/client/parts/bodygroup.lua
@@ -8,8 +8,12 @@ pac.StartStorableVars()
 	pac.GetSet(PART, "ModelIndex", 0)
 pac.EndStorableVars()
 
+function PART:GetNiceBodyGroupName()
+	return self:GetBodyGroupName():gsub("^(%d+)",""):Trim()
+end
+
 function PART:OnShow()
-	self:SetBodyGroupName(self:GetBodyGroupName())
+	self:SetBodyGroupName(self:GetNiceBodyGroupName())
 end
 
 function PART:SetBodyGroupName(str)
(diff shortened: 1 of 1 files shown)


CapsAdmin:
From the looks of it all you've done is replaced the :lower() call with :gsub("^(%d+)",""):Trim(). 

First of all bodygroups are broken? And how does removing starting digits and trim the body group name fix bodygroups? Since you are no longer lowering the name this might break some outfits too.

TW1STaL1CKY:
When a bodygroup part was used, only the owner of the outfit would see the bodygroup being changed. I even tested it before making changes.

When looking at the bodygroup part, I noticed that the owner of the outfit saw the bodygroup name as "headcrab1", if the model was the zombie playermodel. Other players would get something like "9712638 headcrab1" so it wasn't be able to find the right bodygroup.

I put :lower() on 'v.name' as it looks for the correct bodygroup by the way, without :lower() it wouldn't have worked.

CapsAdmin:
Alright. This is an issue with some old stupid hack I did in order to not break older outfits when introducing the new unique id system. This hack adds some unique id in front of the name for other clients but since this was a long time ago this hack can probably safely be removed.

https://github.com/CapsAdmin/pac3/blob/3764ff5c7939ad560e2317239f8d1ff84e30d692/lua/pac3/core/client/base_part.lua#L630-L641

TW1STaL1CKY:
So my 'fix' isn't needed?

CapsAdmin:
I would rather remove the hack or at least blacklist bodygroup name in the copy function instead.

TW1STaL1CKY:
Alrightyy

PR #214, opened 2015-09-28
Change tpose sequence from ragdoll to reference

LUModder:
Affects some custom playermodels (if using the proportions method)


LUModder:
ragdoll

Compared to

CapsAdmin:
wouldn't this break previous outfits?

LUModder:
How so? It's just changing the sequence used which for most default
playermodels is the same as ragdoll.

CapsAdmin:
nvm i thought this was something entirely different

PR #218, opened 2015-10-04
Set hands viewmodel texture to invisible

LUModder:
Instead of setting the position so far out, set the viewmodel texture to invisible so you can still use viewmodel PACs on it.


CapsAdmin:
This was some hack python1320 did that I just left in when porting hands from metastruct to pac3. I think it was because it kept appearing somehow.

PR #240, opened 2015-12-25
!ignorepac command for ULX

LUModder:
Many people have requested it so it can be used for other servers.


CapsAdmin:
its not something that belongs in pac. there should be a gui list or something instead that does not rely on ulx

PR #265, opened 2016-02-26
Fix pace.Restore/DisableExternalHooks not handling entity hooks properly

UselessGhost:
Checks that the hook identifier is a string before running substring on it


UselessGhost:
Even though I doubt the significance of changing isstring to type, I updated it anyways.

CapsAdmin:
When is name never a string?

CapsAdmin:
Oh I see now.

PR #268, opened 2016-03-05
Add workshop part

suchipi:
Adds a part that, given a workshop item id, downloads that workshop content into the cache and mounts it into the client, then updates the model property of all parts in the costume (so that the newly mounted content is used).

Only functions if the server has `pac_sv_workshop_enabled` set to 1 (defaults to 0) and the client has `pac_workshop_enabled` set to 1 (defaults to 1).

### DIFF
--- a/lua/pac3/core/client/parts/workshop.lua
+++ b/lua/pac3/core/client/parts/workshop.lua
@@ -1,0 +1,73 @@
+local PART = {}
+
+PART.ClassName = "workshop"
+PART.NonPhysical = true
+
+pac.StartStorableVars()
+	pac.GetSet(PART, "WorkshopId", "")
+pac.EndStorableVars()
+
+function PART:GetNiceName()
+	if self.WorkshopId and string.len(self.WorkshopId) > 0 then
+		return "workshop " .. self.WorkshopId
+	else
+		return "workshop"
+	end
+end
+
+function PART:OnShow()
+	self:Fetch()
+end
+
+function PART:SetWorkshopId(id)
+	self.WorkshopId = id
+	self:Fetch()
+end
+
+-- If the client wants to download and mount workshop content
+local cl_cvar = CreateClientConVar("pac_workshop_enabled", "1")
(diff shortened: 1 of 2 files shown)


CapsAdmin:
I think this needs to be a separate addon. It's especially pointless being default off, but I understand why it needs to be too.

By the way local foo

suchipi:
It isn't always the local player whose parts need to have their models updated; it's often another player. So that's why I iterate over the parts relative to the part rather than assuming the parts belong to the local player

I have to define the local variable separately for those two functions because they call themselves recursively; something like `local a = b()` doesn't set `a` into the local scope until `b()` finishes, so if `b()` tries to access `a`, it isn't closed over yet.

Should I pull the changes out into a separate addon, then?

CapsAdmin:
I'm saying iterating over the parts owned by the player is effectively the same as what you're doing now. (iterate over `pac.UniqueIDParts[ply:UniqueID()]`) but 
I don't see how your code would affect other players parts, and I don't think it should (since you could override someone else's models with a conflicting addon) but that's an inherent issue with mounting workshop addons.

And that's how it works.

```lua
local function findRootPart(part)
	if part:HasParent() then
		return findRootPart(part:GetParent())
	else
		return part
	end
end
```

findRootPart is accessible inside findRootPart

you can find some more info about it here https://coronalabs.com/blog/2011/09/21/tutorial-scopes-for-functions/

LUModder:
At least this can be disabled and it's not forced

suchipi:
Misunderstood what you meant regarding iterating over player parts; I thought you were suggesting I iterate over `pac.GetParts(true)`. Didn't know `pac.UniqueIDParts` existed. I agree changing it to loop over the part owner's parts is a lot cleaner. I made that change.

thegrb93:
No part limit, file size limit, or checks to prevent overwriting existing models/materials?

LUModder:
File size limit could work, since Outfitter already has that.

PR #274, opened 2016-03-22
GetNoDraw() for hiding PAC when player is not drawn

00875:
This should resolve pain for many coders when they are using ply:SetNoDraw(true) (for example, spectator or ghost, or when player enters vehicle what hides player).
Tested on our server for a long time.

### DIFF
--- a/lua/pac3/core/client/drawing.lua
+++ b/lua/pac3/core/client/drawing.lua
@@ -419,6 +419,11 @@ function pac.PostDrawOpaqueRenderables(drawdepth,drawing_skybox)
 					hide_parts(ent)
 					continue
 				end
+				
+				if ent:GetNoDraw() then
+					hide_parts(ent)
+					continue
+				end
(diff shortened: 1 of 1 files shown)


LUModder:
yes thank god now i can go into hiding on meta without clearing my pac

CapsAdmin:
doesn't this return false under some circumstances on client?

00875:
??? SetNoDraw is networked with server.

CapsAdmin:
I vaguely remember trying this and there was something GetNoDraw that didn't work as expected. I'll revert this if it appears to be a problem I guess.

00875:
I put GetNoDraw about 2 months ago on my server, all working as expected (the problem was when player dies, he is becoming ghost with PAC flying around)

CapsAdmin:
right so when you die the outfit should attach to the ragdoll but this might break that

PR #287, opened 2016-05-10
Mute webstreams when game window is not focused

roboderpy:


LUModder:
Why force it? Why not hook it into "Mute audio in background" cvar

00875:
??? Where that CVar?

LUModder:
snd_mute_losefocus

PR #292, opened 2016-06-06
Revert "Better material browser."

Python1320:
Reverts CapsAdmin/pac3#290

"Flex: I just now found out the mat browser doesnt work"


roboderpy:
wat

LUModder:
Wasn't getting registered for some odd reason, will work it out in the next few days.

PR #340, opened 2016-11-11
Overhaul of the output when converting to E2 holograms.

DatAmazingCheese:
Output turns from this:

To this:

### DIFF
--- a/lua/pac3/editor/client/tools.lua
+++ b/lua/pac3/editor/client/tools.lua
@@ -276,84 +276,64 @@ do
 
 end
 
-pace.AddTool(L"convert to expression2 holo", function(part)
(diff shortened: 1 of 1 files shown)


LUModder:
Why make a new PR?

DatAmazingCheese:
I screwed up changes, new to github. Dunno what I'm doing, should be able to see changes now.

LUModder:
You can edit existing pull requests as long as they're not merged. If you're going to regularly contribute I'd recommend reading up on using git and if you aren't already using a text editor that has plugin support (not N++), doing so. Such being VSCode, Sublime Text and Atom. Not trying to discourage, just trying to point in the right direction.

PR #348, opened 2016-12-17
Merge December

Python1320:
Does everything look good?


LUModder:
Do we HAVE to have the, no offense, ugly icon globally?

CapsAdmin:
It seems alright but some commits mention tweaks which I'm vary of. Although I don't see anything that would be bad at a quick glance.

CapsAdmin:
@LUModder Yes.

PR #373, opened 2017-01-27
Don't draw flashlight scene if flashlight is disabled

roboderpy:
This saves FPS and fixes bug related to "phantom flashlight" on PAC parts


CapsAdmin:
What about lamps?

roboderpy:
https://github.com/Facepunch/garrysmod-issues/issues/3029

PR #417, opened 2017-06-08
for what i've seen this works, and prevents error on part init

Earu:


CapsAdmin:
What's the error? With the way Lua works this shouldn't make a difference.

Earu:
@CapsAdmin now it should be fine, and not error if the part is getting initalized before the client has fully initialized.

Earu:
Agreement that fix needs to be global and not ductape

CapsAdmin:
but I'm lazy atm so it's better than nothing maybe

PR #425, opened 2017-06-21
Call PACResetBones hook on bone reset

roboderpy:
this will ease the pain of coders who rely on player bones in their addons


roboderpy:
When this would get merged i will remove `ResetBones` override in my addon

PR #434, opened 2017-06-28
Menu improvements

roboderpy:
Make `Copy` works like `Clone`, but with auto parent to selected part
Don't populate `Backups` after `Load` was hovered, and moved `Backups` to bottom of menu
Hide `DFrame` useless panels in PAC3 Editor Panel


CapsAdmin:
One "feature" of the dframe buttons is that when you click in the upper right corner of the editor it will close it. It can probably be added back somehow in a better way.

Dunno how many people used this, maybe just me.

PR #441, opened 2017-07-03
Re-Add close button[2]

roboderpy:
It now doesn't hide and doesn't call MakePopup(); so code of any quality now should work with it (no more menu closing on MetaStruct); And enabled old close button, but made it invisible

### DIFF
--- a/lua/pac3/editor/client/panels/editor.lua
+++ b/lua/pac3/editor/client/panels/editor.lua
@@ -124,20 +124,13 @@ function PANEL:Think(...)
 	end
 
 	if self.pac3CloseButton:IsValid() then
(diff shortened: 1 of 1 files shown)


Datamats:
I think you are looking for http://i.imgur.com/ctCyxmQ.png

Earu:
Please be aware that you're not better than everyone.

roboderpy:
Caps said he is using close button on top of window, rather than using menu. So i added close button that anyone will be aware of

CapsAdmin:
@Earu what

@roboderpy I thought I was the only one who used it but some other people started asking where it went as well.

PR #471, opened 2017-08-27
Events API

roboderpy:
For modders who want to add custom events in PAC3

Will be merged after more tests.


roboderpy:
Seems stable

PR #479, opened 2017-09-03
Fix flickering / improve draw suppression

TheFreezebug:
The most hax, ever.

Fixes the frame counting / suppression of PostDrawOpaqueRenderable calls. The frame suppression no longer depends on how many draw calls are in the current frame, but matters that the render stack executed in a correct order (There's no reason why it shouldn't). 

This fixes outfit flickering. 

There's a pull request in on the master repo of GMod to hopefully make this less hacky .

https://github.com/Facepunch/garrysmod/pull/1406

### DIFF
--- a/lua/pac3/core/client/drawing.lua
+++ b/lua/pac3/core/client/drawing.lua
@@ -509,7 +509,9 @@ do
 
 	function pac.NeedsDepthPass()
 		last_DOR_Window = true -- sets the flag to notify the postdrawopaquerenderables that this is our last opportunity to draw. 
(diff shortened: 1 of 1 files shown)


CapsAdmin:
in the develop branch i removed all of this hack stuff and i'm just using RenderScreenspaceEffects with cam.Start/End3D()

that seems to work fine here

TheFreezebug:
🎊 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎊

Issue #523, opened 2017-11-06
issue with legacy Entity and the new Entity2 parts, they cause GetModel() to go out-of-sync between the Client and the Server


CapsAdmin:
Hazbelll - 10/13/2017
Also another issue with legacy Entity and the new Entity2 parts, they cause GetModel() to go out-of-sync between the Client and the Server.

How to reproduce:

1. Add an Entity part to PAC.
2. Change the model field to, for example: models/props_c17/furnitureStove001a.mdl
3. Delete the PAC and your Player Model should revert to its usual model.
4. Print the model using Entity:GetModel() on the Server. It should return the current model correctly.
5. Print the model using Entity:GetModel() on the Client. It should not return the current model, but the one you set via PAC.

roboderpy:
Entity2 just causes model desync in general. Model changes on all clients, but the server stays with same model causing top notch both visual and logical issues

roboderpy:
![aaaaaAAAAAAAAAAAA](https://i.dbot.serealia.ca/2017/11/92ed3185b6_hl2_2017-11-10_17-36-00.png)

thegrb93:
I tried reproducing this issue but couldn't. One thing I think should change though is deleting the entity part should revert the player's model to normal.

thegrb93:
I found whenever an entity with custom model enters PVS and triggers 'NotifyShouldTransmit', the model that is rendered goes back to whatever the server's value is, however GetModel() will still return the set custom model.

Further, SetModel again on the entity to return it to the custom model will not work since the lua state thinks it's already that model (even though the original model is being rendered).

thegrb93:
When the server's model modifier is enabled, it doesn't seem to be a problem with players since it sets their model serverside.

thegrb93:
Another issue is removing the Entity part will not clear whatever custom model the player or entity is set to.

roboderpy:
this was fixed in two commits before, both from me and @thegrb93

PR #559, opened 2018-01-08
Preserve entity model after PVS update and reset model when entity part is removed.

thegrb93:
Probably not the best way.

Is there a part hook for when an entity part's owner is changed? I think that should be used instead of OnShow/OnRemove


thegrb93:
Should fix https://github.com/CapsAdmin/pac3/issues/523

CapsAdmin:
I can cleanup later. Thanks.

thegrb93:
Once you do, I can do the same for the original Entity part, or you can, whichever is fine.

Issue #560, opened 2018-01-09
Bass starts playing at full volume sometimes, then returns to normal when entering the play radius.


thegrb93:

thegrb93:
That's my theory anyway. I'll see if I can figure it out.

thegrb93:
seems to be the webaudio part

thegrb93:
maybe it's because OnDraw is being called with pos at LocalPlayer() pos due to the outfit being outside PVS?

roboderpy:
Yep because of bad BASS integration. But no, because of pursuit state of bass channel. Sound starts to play before callback is called (on low framerate of client), which causing it to be heard at maximum value because callback turns it off.

thegrb93:
Shouldn't the noplay flag prevent that?

roboderpy:
![bark](https://i.dbot.serealia.ca/2018/01/346fd9c1c3_firefox_2018-01-09_13-55-48.png)

thegrb93:
Can you report to garrysmod-issues if you haven't yet? I think you know more about it than I do.

thegrb93:
These are the flags used, `local flags = "3d noplay noblock"`

roboderpy:
im lazy, and they won't fix it. issue will just get rotten in place

thegrb93:
noblock means it's async right? could removing that flag fix it?

roboderpy:
I was getting loud sound in both cases, with `noplay` and `noblock` enabled/disabled. it is just an issue with bass. Only way to workaround this is writting file into sound/ folder and playing it using `EmitSound` which uses BASS for oggs too, but it doesn't seem to have any issues with playing the sound.

thegrb93:
Are you sure this is a problem with BASS? I think it's a problem with the outfit playing the sound to be outside PVS.

thegrb93:
It happens more when that is the case.

roboderpy:
```
[PAC3] failed to render Player [2][MAGKEKV3000 (null98932)]:
[VLL: pac3-original - pac3/core/client/parts/webaudio.lua]:172: attempt to index local 'stream' (a nil value)
```

roboderpy:
Also `3d noplay noblock` is not working for me at all ![lol](https://i.dbot.serealia.ca/2018/01/d149c56418_hl2_2018-01-14_08-55-39.png)

roboderpy:
all should be working now

PR #561, opened 2018-01-10
Fixes to webaudio

thegrb93:
Fix webaudio sounds playing before their position is set. Also make webaudio sounds play after they're loaded if they should be playing.


thegrb93:
Fixes https://github.com/CapsAdmin/pac3/issues/560 hopefully

thegrb93:
I take it everyone has at least looked at it. Should be safe to put on develop branch.

PR #562, opened 2018-01-12
maybe fix serverside playerscaling

Techbot121:
it got renamed to pacx I believe.


CapsAdmin:
Yeah but you should check for its existence (if pacx then) since it's optional and may be removed in certain non sandbox gamemodes.

PR #569, opened 2018-01-24
Smallish Fixes and exposing the bone selector hooks to the outside for external rendering

Kamshak:


roboderpy:
*A bit* unstable changes, merged into develop

PR #580, opened 2018-02-17
Fix issue w/ RenderView

devSera:
RenderView was causing PAC Rendering to be fucky

### DIFF
--- a/lua/pac3/core/client/part_pool.lua
+++ b/lua/pac3/core/client/part_pool.lua
@@ -710,7 +710,7 @@ do -- drawing
 							(ent == pac.LocalPlayer and ent:ShouldDrawLocalPlayer() or (ent.pac_camera and ent.pac_camera:IsValid())) or
 							ent ~= pac.LocalPlayer and
 							(
-								((fovoverride ~= 0 or util_PixelVisible(ent:EyePos(), radius, ent.pac_pixvis) ~= 0) or (dst < radius * 1.25)) and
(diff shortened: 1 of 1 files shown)


roboderpy:
Pac doesn't draw itself when `RenderView` is called unless `pac_suppress_frames` is set to 0
Also i didn't found any issues with **poper** usage of `RenderView`.
If PAC3 *getting fucky* you are calling `RenderView` in a wrong hook and thus you are breaking render priorities and stuff.

devSera:
RenderView makes http://wiki.garrysmod.com/page/util/PixelVisible break

roboderpy:
Yes because you are calling `RenderView` before actual scene draw. Also many other stuff breaks, like `Vector:ToScreen()`, but that's because Source Engine doesn't reset current scene properties when it pops current screne context using `cam.End3D`

Breakage of pac and other stuff is still coder's fault. He should call `RenderView` at least in `PostDrawHUD`.

If Caps would approve this PR, he could merge it. I see no point in extra hack because of somebody's lack of understanding how to use `RenderView`.

Tenrys:
I don't see why you wouldn't merge it, if it causes no extra issues and is a convenience.

roboderpy:
After some tests see no issues with it. if it wouldn't cause potato PC to choke thats good.

devSera:
Why not just solve the problem on PAC's side so it's not relying on

CapsAdmin:
Wouldn't this cause the pixel visibility to be pointless? If that test fails then it checks with IsLineOfSightVisible

I'm not sure if the pixel visible stuff works in RenderView, but there are hacks in pac that prevents them from being drawn more than once on purpose. I think you can can turn this off programmatically (can't check atm) if really needed.

Issue #571, opened 2018-01-28
Bonemerged model keeps glitching out of place


spaceboat:
So im trying to bonemerge a jacket onto my character but when I move to certain places, the bonemerged part of the jacket moves out of place from where I want it to stay. Not sure what im doing wrong

spaceboat:
Yeah, like if i walk to the right it starts moving out of place

thegrb93:
But does it stay that way while holding still? or does it return to the correct position?

spaceboat:
Yeah it stays in the position but if i look at a certain angle it goes back to place

thegrb93:
Could you upload your pac?

spaceboat:
Alright, sending one now

thegrb93:
I think I've found the issue to be due to FollowPart code which adds euler angles instead of using proper rotation matrices. I'll fix it.

thegrb93:
@CapsAdmin @roboderpy I changed the code of FollowPart in bone.lua to this, which seems to fix the gimble-locking/angle addition artifacts, but it screws up pre-existing pacs that use FollowPart. Idk if it's right, mind looking?

```
local part_ang, follow_ang = Matrix(), Matrix()
part_ang:SetAngles(part.Angles + part.AngleOffset)
follow_ang:SetAngles(part.FollowPart.cached_ang)
local angles = (part_ang * follow_ang):GetAngles()

if part.FollowAnglesOnly then
	local pos = mat:GetTranslation()
	mat:SetAngles(angles)
	mat:SetTranslation(pos)
else
	mat:SetAngles(angles)
	mat:SetTranslation(part.Position + part.PositionOffset + part.FollowPart.cached_pos)
end
```

thegrb93:
Found some neater code. Produces the same results
```
if part.FollowAnglesOnly then
	local pos, angles = LocalToWorld(Vector(), part.Angles + part.AngleOffset, Vector(), part.FollowPart.cached_ang)
	mat:SetAngles(angles)
else
	local pos, angles = LocalToWorld(part.Position + part.PositionOffset, part.Angles + part.AngleOffset, part.FollowPart.cached_pos, part.FollowPart.cached_ang)
	mat:SetTranslation(pos)
	mat:SetAngles(angles)
end
```

thegrb93:
Made some neater code that produces the same results
```
if part.FollowAnglesOnly then
[...]

PR #583, opened 2018-02-18
Fix followpart using euler angle addition for following bones.

thegrb93:
This is going to break a lot of pacs that use followpart but they were technically already broken. I'd make sure this is the behavior we want before merging though.

Fixes https://github.com/CapsAdmin/pac3/issues/571

### DIFF
--- a/lua/pac3/core/client/parts/bone.lua
+++ b/lua/pac3/core/client/parts/bone.lua
@@ -168,13 +168,12 @@ function pac.build_bone_callback(ent)
 				local mat = ent:GetBoneMatrix(data.bone)
 				if mat then
 					if part.FollowPart:IsValid() then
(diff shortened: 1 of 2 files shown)


Techbot121:
how about deprecating the old one but still make it functional? I think the scale changes were done that way too.

CapsAdmin:
Yes, just make this a new property and deprecate the old one.

We can also think about making "bone2" if needed as well

thegrb93:
What should the new field be called? FollowPart2, FollowPartNew, FollowPartFixed, FollowPartMatrix?

CapsAdmin:
Can you make a new part called bone2.lua with the fixes and put it under the pac4 submenu?

There are other things I want to change about bones I think that might break everything.

thegrb93:
All right

thegrb93:
Should be ready

PR #586, opened 2018-02-22
Turn off blend-override until its fixed

thegrb93:
https://github.com/Facepunch/garrysmod-issues/issues/3396


thegrb93:
```
>lua_run_cl hook.Add("PostDrawOpaqueRenderables","A",function() render.OverrideBlendFunc(true, BLEND_ONE, BLEND_ONE) render.OverrideBlendFunc(false) end)
```
You'll see the city material in gm_flatgrass blend improperly, so the issue is not specific to translucent parts.

PR #605, opened 2018-03-15
fixed the lagger exploit

joeyjumper94:
before the change, people can lag a server by spamming the "pac_submit" net message, they'd send this message to the server tens of thousands of times in a frame every half second.
now we make it so the function can be executed only once per frame, per player. i've tested it on pony withers and it does work

### DIFF
--- a/lua/pac3/editor/server/wear.lua
+++ b/lua/pac3/editor/server/wear.lua
@@ -300,7 +300,11 @@ end
 
 util.AddNetworkString("pac_submit")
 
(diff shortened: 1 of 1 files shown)


CapsAdmin:
This could potentially make it so you can't wear an outfit with 5 groups though.

joeyjumper94:
scrapping this edit cause i've had only partial success in making a workaround that spreads the multiple net messages across multiple frames. stay tuned cause there is another method that also works and allows multiple groups.

PR #606, opened 2018-03-15
v2 of the fix

joeyjumper94:
due to bad networking practices the method of limiting a player to one net message per frame would block PAC3s that have multiple groups from loading correctly.
this method checks to see if they sent more that 100 in a 0.1 second time frame


CapsAdmin:
I don't have access to a computer right now but you should store a time or something instead of spamming timer.Simple.

It would be very much like your previous PR but with an additional counter that resets to 0 when okay.

PR #659, opened 2018-06-07
Fixed PAC-to-Holo holoclip support

CheezusChrust:
I got holoclips working in the PAC-to-Holo converter tool. Tested 1 model with 1 clip, multiple models with 1 clip each, 1 model with multiple clips, and multiple models with multiple clips each. All cases seem to work.


roboderpy:
merged to develop

PR #715, opened 2018-10-27
obsfucate part_pool.lua at runtime

CapsAdmin:
this obfuscates and randomizes upvalues

### DIFF
--- a/lua/pac3/core/client/init.lua
+++ b/lua/pac3/core/client/init.lua
@@ -80,7 +80,92 @@ include("pac3/libraries/webaudio/streams.lua")
 pac.animations = include("pac3/libraries/animations.lua")
 
 include("parts.lua")
-include("part_pool.lua")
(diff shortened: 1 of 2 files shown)


CapsAdmin:
@Python1320 @thegrb93 @Kamshak

thegrb93:
ack, this upsets my server's anticheat. I'll see if I can make an exception or something

CapsAdmin:
Good point about the uid thing. I do prepend _ in front of it in all places except one where I forgot, but it should probably be part of gen_uid and then the function being renamed to generate_identifier or something.

Is it being flagged because of runstring?

thegrb93:
yea hooks added in code that has not been AddCSLuaFile'd or something like that get flagged as cheater's hooks.

thegrb93:
I can just add an exception to it.

PR #723, opened 2018-11-06
Cleaning up and implementing new features to hologram conversion

DatAmazingCheese:
Allows people to set hologram positions relative to others. Unable to do so automatically but it is extremely easy to do so now.

In order to do so automatically would need rewrite or further understanding to how pac handles part indexes.

Cleanup to conversion and formatting removing redundant functions.

Small fixes to spelling errors on my part and small inconsistencies with code.


DatAmazingCheese:
Need to double check some implementation of systems and explaination of systems.

PR #724, opened 2018-11-07
Implementation of new features to parts > holograms + cleanups to tools.lua

DatAmazingCheese:
Implemented new features to parts > holograms such as separation of local entity and parent entity. Added documentation. Cleaned up tools.lua for consistency.

Questions regarding line 200 on tools.lua with the tool "round numbers" and its unused functions. Are they used for anything? Does this tool still work as it seems to not work.


thegrb93:
Is this working? I've never tried it since I don't use e2 often.

DatAmazingCheese:
> 
> 
> Is this working? I've never tried it since I don't use e2 often.

Yes, this is working.

Issue #722, opened 2018-11-05
Bone HideMesh/InverseHideMesh causes crashes by scaling bones to inf (Temporarily fixed)


Finalization:
Not sure if its an entity or lighting issue on the server or my pac3. But every time I die with one of these pacs I have a very high chance of crashing myself or anyone near and they don't even have to be looking at me. If you guys could help me out that would be great.

These ones don't crash:
https://pastebin.com/sxD6Lxjr 
https://pastebin.com/CKkXPcHH
This ones do crash:
https://pastebin.com/S0ZNSSHS
https://pastebin.com/67YQ51sE

If anyone could help me out that would be great, I don't want to get accidently perm banned for crashing a server. Please and thank you.

Jophes:
Looks like this still reproduces on the latest develop branch and gmod build,

Took a look at the mdmp but it doesn't really help much without the correct symbols, except to know that it crashed in vphysics.dll

`Unhandled exception at 0x6539CCBC (vphysics.dll) in hl2_181215_crash_2018_12_15T18_30_10C0.mdmp: 0xC0000005: Access violation reading location 0x082ADBF4.`

Through trial and error, it looks like it this is caused by hiding bone meshes using a bone part. The way the bone parts hide bones, is by setting their scale to infinity, or rather lua's infinity `math.huge`, like so:
`scale = Vector(math.huge, math.huge, math.huge)`
If you change these constants to 0, the game will no longer crash upon dying. However, it will no longer correctly hide the other parts of the body, as the player sort of turns into a stick man.

I believe this to be an issue with the way the Physics engine is handling Infinity from Lua, which some how becomes an issue when the player becomes a rag doll. I was going to try changing the bones just before the player dies, but the hooks for the player dying are all server side. 

Not really sure how to go about fixing this while keeping all of the current functionality, will have another look tomorrow.

CapsAdmin:
this should finally be fixed

PR #742, opened 2018-12-16
Disable Bone HideMesh functionality to prevent crashes

Jophes:
Disables the functionality of the hide mesh property of the bone part, until a complete fix can be found. 

When this hide mesh property is enabled on the bone part, if the player dies, depending on the bone its linked to, it can cause the physics engine to throw unhandled exceptions, completely hang the hl2 process, and other undefined behaviors. This could be very destructive as players can enter servers with Pac3 and cause all players nearby to crash, with minimal effort.

This is caused by setting the bone scale to scale_inf which is a vector with math.huge in x,y and z. This somehow becomes a problem for the engine when the player dies. It may be possible to reset the scale of the bones just before the player is turned into a rag doll but I haven't been able to come up with a solution that works other then simply disabling the property entirely. 

This fixes issue [#722](https://github.com/CapsAdmin/pac3/issues/722)


thegrb93:
Thanks for narrowing down the cause. We'll keep the issue open and rename it so that it stays known.

Issue #756, opened 2019-02-03
Numerically named bones ignore preceding zeros


lilggamegenius:
Basically bone parts that are set to bones that start with the number 0 seem to reset and map themselves to a non-existent bone due to the deletion of said 0.. Basically whenever it wears, it tries to get rid of the 0 at the start of the bone name, thus making the bones impossible to use.
![image](https://user-images.githubusercontent.com/6175821/52170708-f1bc7600-2714-11e9-828d-71cc6ef57df5.png)
All of these bones are basically 00, 01, 02, and because of that it tries to default said bone parts to searching for 1, 2, 3
I had a whole PAC break because of this issue tbh, I can't even wear the thing anymore without it combusting entirely.

_Issue found by LogZombie_

thegrb93:
Should be fixed. Let us know if it isn't and we'll reopen.

PR #757, opened 2019-02-05
Prevent 0 prefixed strings becoming numbers

Jophes:
Fixes issue #756 

When strings which are also valid integers are used as names, they are converted to TYPE_NUMBER_UID mistakenly, causing string properties to be converted into numbers

The fix prevents any strings that are longer than 1 character and begin with 0, from being converted


thegrb93:
Lgtm

Issue #749, opened 2019-01-07
Flex not working on models other than base playermodel (eg.: flexing a second head not working) 


Kenajcrap:
First off, my apologies if I'm misunderstanding the limitations of flex and this was actually never possible or I'm just not doing this right

If that's not the case, here is the problem: I created a model, set it's path to breen and was trying to flex it's face, but the face being flexed was the main playermodel instead.

Here is what I've already tried and the result:

Created a flex as a child of the target model >> main playermodel flexes instead
Created a entity and a flex as a childs of the target model >> main playermodel flexes instead
Created a entity as a child of the target model and created a flex as a child of the entity >> nothing flexes?

I'm on mobile, so I will only be able to upload screenshots in a couple hours, until then this might be enough just to know if this is an actual problem or just not possible.

Ty in advance for your time.

bluemetaknight:
Found a solution. Just untick root owner on the flex part and it should work.

Jophes:
Just to clarify for Kenajcrap, this was the intended functionality. 
When root owner is ticked, the flex part alters the player instead of the part's parent. 
I submitted PR #760 to change the default value to false and add a description when you hover over the property to make this clearer.

PR #760, opened 2019-02-11
Improved description of flex property RootOwner

Jophes:
Added a description to the RootOwner property, to prevent confusion
Changed the default value of RootOwner to false, to match the functionality of other parts

This is to help prevent confusion as seen with issue #749

### DIFF
--- a/lua/pac3/core/client/parts/flex.lua
+++ b/lua/pac3/core/client/parts/flex.lua
@@ -19,7 +19,7 @@ pac.StartStorableVars()
 	})
 
 	pac.GetSet(PART, "Weight", 0)
-	pac.GetSet(PART, "RootOwner", true)
+	pac.GetSet(PART, "RootOwner", false, { description = "Target the local player instead of the part's parent" })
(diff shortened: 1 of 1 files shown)


thegrb93:
I'm pretty sure it won't, but just to be sure, this won't change everyone's pacs that use that part and use the default value, right?

Jophes:
Hadn't thought of that, thanks

Just checked it by saving a pac that used the default of true, 
Then tried loading it with a default of false, the property was still set to true
Should be good

CapsAdmin:
In the past we used to not store default values in outfits, but now it stores everything I think. This makes changing default values like this possible.

thegrb93:
So old old outfits will change if they use this part. They should be easy to fix at least.

Jophes:
Would it be a good idea to add versions to saves, so that providing backwards compatibility with older saves is easier in the future?

thegrb93:
I'd say we should only if the entire saving system changed. Small part changes and backwards compatibility just leads to bloat.

PR #763, opened 2019-02-17
Not supposed to be 'length'?

FarukGamer:
During my digging trough the Logs i have discovered this error.

```
[ERROR] lua/includes/extensions/string.lua:261: attempt to compare nil with number
1. NiceSize - lua/includes/extensions/string.lua:261
2. unknown - addons/pac3_dev/lua/pac3/core/shared/http.lua:108
```
'len' is not existing, so it must be 'length'?


thegrb93:
Looks correct

PR #779, opened 2019-05-02
Some changes to shortcuts

FarukGamer:
Make some shortcuts available only if the editor is in the foreground,
This should prevent e.g. the accidental saving during the free view.


thegrb93:
Looks good. One question, will the editor still have focus when manipulating a part with the manipulators outside of the editor?

FarukGamer:
manipulating parts with the axis works in my test.
https://streamable.com/s/ue17k/jrjqsn

FarukGamer:
manipulating parts with the axis works in my test. ( holding ctrl down )
https://streamable.com/s/ue17k/jrjqsn

PR #790, opened 2019-05-21
Add rotation/position offset to e2 converter. Also optimize gsub and fix GetSkin lua error

thegrb93:


thegrb93:
@megalatapalapapuas see if this fixes it

thegrb93:
@megalatapalapapuas now it should work

PR #797, opened 2019-06-01
Made the FOV/zoom slider actually visible.

Yagira:
Should help the 2 people in the community that actually use this feature. :)

### DIFF
--- a/lua/pac3/editor/client/panels/editor.lua
+++ b/lua/pac3/editor/client/panels/editor.lua
@@ -48,11 +48,19 @@ function PANEL:Init()
 	self.exit_button.Paint = function(self, w, h) derma.SkinHook("Paint", "WindowCloseButton", self, w, h) end
 	self.exit_button:SetSize(31, 26)
 
-	self.zoom = vgui.Create("DSlider")
-	self.zoom:SetLockY(0)
-	self.zoom:SetSize(100, 20)
-	self.zoom:SetSlideX(1)
-	--self.zoom:SetVisible(false)
+	self.zoomframe = vgui.Create( "DPanel" )
+	self.zoomframe:SetSize( 180, 20 )
+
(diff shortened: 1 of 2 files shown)


Yagira:
Looks like this in-game.
image

thegrb93:
Lgtm. I'm not sure why pingu wanted to make another version but this one came first and matches the style better. I guess if you want to tweak it more, @pingu7867 , you can do another PR on top of this one.

pingu7867:
Here's why I chose what I chose.
On one hand, an invisible slider whose pin is half out of the screen was not right.
On the other hand, Yagira's version has some issues, namely that you can't move it up like you can move my window.
So I thought a small window was the best thing in terms of choice. You can close it if you don't need the zoom slider. You can move it around anywhere in the screen to access it easier.

But Yagira's version is independent from mine, and he revealed his slider after I started working on mine. I didn't want to make _another_ version, I simply wanted to make _a_ version for visuals, because Sputnik asked for it in the discord. I also wanted to know which one people preferred, so I gave it a shot.

Since you suggest it, I'll see if there's things to tweak. But we need the community's feedback for what they want.

But other than that, I want to know what's the guidelines for how I should edit code, knowing that I'm not a lua pro and that, as we've seen, there's different interests that conflict about what direction to take pac when it comes to stuff like the user interface. How to weigh the side of the cutting edge pac users who adapt and may welcome some changes, but on the other side the general community who don't always update but who eventually install the dev branch, and who might have problems with design changes? Since we get into that philosophical area, maybe DM me on discord for further discussion.

thegrb93:
Just do whatever you think is right and we'll give feedback. A comprehensive guideline isn't really feasible.

thegrb93:
Nothing is set in stone, and there's always room for improvement so don't get discouraged.

PR #798, opened 2019-06-01
A DFrame window to contain the zoom slider (Alternative to Yagira's proposed slider)

pingu7867:
It's not good that the zoom slider is just invisible in the corner, and even the rare few who know it exists have difficulty because of how broken its positioning is (the slider pin is half out of the screen, and it is also not easy to target because it is invisible and small). A window has some advantages over a slider that is always at the bottom of the screen.

### DIFF
--- a/lua/pac3/editor/client/panels/editor.lua
+++ b/lua/pac3/editor/client/panels/editor.lua
@@ -49,10 +49,16 @@ function PANEL:Init()
 	self.exit_button:SetSize(31, 26)
 
 	self.zoom = vgui.Create("DSlider")
-	self.zoom:SetLockY(0)
-	self.zoom:SetSize(100, 20)
+	self.zoom:SetLockY(0.5)
+	self.zoom:SetSize(200, 20)
 	self.zoom:SetSlideX(1)
 	--self.zoom:SetVisible(false)
+	self.zoomframe = vgui.Create( "DFrame" )
+	
+	self.zoomframe:SetSize(220,50)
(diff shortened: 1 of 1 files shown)


pingu7867:
What's the best way to choose the one that's the best for people and the best in code quality?

thegrb93:
Whys the slider not parented to the DFrame?

pingu7867:
I'm very sure that the slider is parented to the DFrame. It's at line 59. Where is the code wrong?

thegrb93:
ah ok. Usually it's in the vgui.Create iirc

thegrb93:
You'll need to resolve conflicts if you want to tweak it, but I don't like the DFrame very much.

pingu7867:
Of course it conflicts. Yagira and I had different ideas of what should be the zoom slider's visual. So I think that whatever people want from this feature is inherently a problem of differing preference, so we can leave tweaking that to each person to choose for themselves by editing pac clientside for singleplayer. I for one will keep using my DFrame tweak.

But I still have my previous question which is what's the community's preference? It's kind of important to decide what we do commit. But it used to be invisible before the recent commit, so it's a bit of a chicken and egg problem. Most people didn't know about it, so it's hard to tell.

But for now Yagira's solution should work, though I don't know why it was chosen to default to a FOV of 45 when the editor is opened. That's a problem. It'll throw off people for sure because it used to be at 90 by default.

thegrb93:
I was a bit thrown off too by the initial fov. We can change that.

PR #803, opened 2019-06-05
Some refract shader changes. 

### LINKED ISSUE #703: Material Refract part non-functional.
Big bad. Idk if this has been posted yet but material refract part under experimental doesn't do anything atm apparently, when put on a model it just turns said model invisible and attempts to tick flags for model, etc. or adding textures to it doesn't do anything either, it remains invisible no matter what.

Yagira:
Fixes #703

### DIFF
--- a/lua/pac3/core/client/parts/material2.lua
+++ b/lua/pac3/core/client/parts/material2.lua
@@ -475,6 +475,7 @@ for shader_name, groups in pairs(shader_params.shaders) do
 						self[property_name] = val
 						local mat = self:GetRawMaterial()
 						mat:SetVector(key, val)
+						if info.recompute then mat:Recompute() end
 					end
 				elseif info.type == "vec4" then
 					-- need vec4 type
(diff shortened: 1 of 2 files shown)


Earu:
Don't worry guys im pushing that! /s

thegrb93:
Looks sane, but I don't have any idea what and how the change affects it.

thegrb93:
The defaults are fine. Any explanation as to what the recompute is for?

Yagira:
The material needs to be recomputed for $vertexcolormodulate to actually update. Previously it was set to (1,1,1) which made it invisible and it never changed unless you also changed something else that triggered the recompute. It caused most people to assume that the part didn't work at all. I guess it doesn't really matter anymore after changing it to (0,0,0) since it makes it work by default and there is no need to change $vertexcolormodulate in the editor at all.

thegrb93:
Yeah I see a few other parameters doing the same. Not sure why there's an option for it. Lgtm

PR #807, opened 2019-06-11
Zoom slider settings and some small fixes

Yagira:
Zoom settings include persistent zoom, zoom on mouse wheel and smooth zooming.


thegrb93:
That's a hell of a lot of config for a zoom slider. I guess I'll check it out tonight.

PR #815, opened 2019-07-04
Added CVar to opt-out of using 'camera' and 'player_eyes' as a bone

mitterdoo:
This lets players disallow outfits from using the local player's `EyePos()` and `EyeAngles()` with the cvar `pac_allow_camera_bone`

### DIFF
--- a/lua/pac3/core/client/bones.lua
+++ b/lua/pac3/core/client/bones.lua
@@ -5,7 +5,7 @@ local Angle = Angle
 local Vector = Vector
 local util_QuickTrace = util.QuickTrace
 local pac = pac
-local pac_isCameraAllowed = pac.CreateClientConVarFast("pac_allow_camera_bone", "1", true, "boolean")
+local pac_isCameraAllowed = pac.CreateClientConVarFast("pac_enable_camera_as_bone", "1", true, "boolean")
 
 pac.BoneNameReplacements =
 {
--- a/lua/pac3/core/client/bones.lua
+++ b/lua/pac3/core/client/bones.lua
@@ -5,6 +5,7 @@ local Angle = Angle
 local Vector = Vector
 local util_QuickTrace = util.QuickTrace
 local pac = pac
+local pac_isCameraAllowed = pac.CreateClientConVarFast("pac_allow_camera_bone", "1", true, "boolean")
 
 pac.BoneNameReplacements =
 {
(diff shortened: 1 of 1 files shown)


thegrb93:
Can you explain the purpose of this and why it's a hidden cvar?

mitterdoo:
Ah, my mistake on "hiding" the cvar; I hardly make any contributions to PAC. There will be an additional commit to add it to user menus.

There have been a few outfits on our server, that display unwanted things on other players' screens. What's worse, is that the viewing player has no way of knowing whose outfit is the cause of the parts shown on their screen, other than excruciating trial-and-error. This is a quick fix for players who don't want to use their current camera as an attachment for other outfits.

However, in the long run, perhaps a "blame" feature could take the place of this quick fix. Maybe a tracker that keeps track of players whose outfits are using the camera/player_eyes bones, where the list is shown when the player runs a command?

Yagira:
Isn't it better to just remove the camera/player_eyes bones? Is there an actual use for them except for trolling other players?

iskawhiskers:
Yes @Yagira you could use it legitimately to make UI elements like health bars

thegrb93:
Oh that explains the time someone made a HUD but it kept showing up on my screen.

thegrb93:
I guess the other question is is this an effective solution? I think better would be to not allow players to attach their parts to others or to at least notify them of the part. Thoughts?

pingu7867:
Another legitimate use is stuff like hallucination abilities (for example I remade some abilities of Nyx from Warframe and one of them is that the enemies around you start to hallucinate an image of Nyx and start attacking each other. Not translatable very much to gmod but that's an example), but even if we can name legitimate uses of it, we still have the value/principle that this requires people to consent to that aspect of "PAC fights" and it also requires that to be used in moderation and work properly and to be targeted properly instead of everyone.
Perhaps who gets to project stuff on others' screens is best handled with a peer to peer whitelisting system. If possible, have a new projectile damage type or some other method that marks a single (or all in a blast radius) player for the projectile's outfit part name to be used in their screen if the player whitelisted you.

Yagira:
Well if its mostly used for HUDs then these should render only for the owner. That way everyone can still use them and it won't annoy other people.

iskawhiskers:
@Yagira My previous comment pertained mostly to UI elements other players are _supposed_ to see, e.g. boss battle healthbars

thegrb93:
I think allowing players to disable it is good. Just need a setting checkbox for it and it'll be ready to merge.

mitterdoo:
I renamed the cvar earlier so it will show up in the PAC settings in the Q menu, so that part's already done

thegrb93:
Ok, I'll test it out

PR #816, opened 2019-07-08
Return correct color values on color proxies.

Yagira:


thegrb93:
Why do experimental parents want 0-1 while not experimental is 0-255? Seems strange to me. The added conversions here will probably break a lot of outfits.

Yagira:
> Why do experimental parents want 0-1 while not experimental is 0-255? Seems strange to me.

You'll probably have to ask caps. 

> The added conversions here will probably break a lot of outfits.

I suggested this change mostly because these proxies don't work "by default" and you have to manually tweak them in the expression. Some of them return 0-1 while others do 0-255 which seems really inconsistent and I think that it causes most inexperienced uses to just assume that the proxy doesn't work at all. While there is a chance for it to break existing outfits, I believe that people who know what they are doing could easily fix them.

thegrb93:
I'll see what they think, if we should edit the parts themselves or if this is more suitable. Either way, outfits will break, but I agree that the current way is inconsistent and confusing.

lua/pac3/core/client/parts/proxy.lua
```diff
@@ -908,6 +960,8 @@ function PART:RunExpression(ExpressionFunc)
 	return pcall(ExpressionFunc)
 end
 
+local oldclass = ""
```
roboderpy: ?

thegrb93: Used to only update the expression if the parent becomes a different class?

thegrb93:
Yah good call

thegrb93:
I'll take a look at the color parameters of the parts and see if there isn't a more elegant solution.

thegrb93:
If the parameters in the editor are 0-1 I think the proxy should be that way too.

PR #819, opened 2019-07-13
Check if the editor is valid

FarukGamer:
Maybe explains by itself... *cough* fix for zoomslider https://github.com/CapsAdmin/pac3/pull/807


thegrb93:
Using the world panel's hook seems like really bad practice too.

PR #836, opened 2019-10-06
change from number to string

### LINKED ISSUE #831: Other player's PAC outfits seem to be failing to load on player's client (Occasional. Somewhat inconsistent)
While not completely sure as to why this happens, the use of the "Request Outfits" feature has almost become essential. Other player's PAC outfits don't seem to be automatically loading upon the player hitting the wear button, resulting in the only method of seeing their PACs to be requesting outfits. 

This doesn't seem like too much of an issue at first, but considering some players switch PACs frequently, and the fact that PAC3 used to be able to load PACs almost instantly to the client once worn serverside.. I'm beginning to believe this is at least a bug since I don't see exactly why it would be removed, unless there's any additional information that I'm missing.

Techbot121:
should fix #831 and other issues related to wearing or filter issues.

Not sure why it was changed to number but UIDs are fairly long and would overflow in some cases.

I tested this on the pac server and it seemed to work fine but in case I missed something please do tell me.

### DIFF
--- a/lua/pac3/editor/server/wear.lua
+++ b/lua/pac3/editor/server/wear.lua
@@ -408,7 +408,7 @@ function pace.RequestOutfits(ply)
 
 			if owner:IsValid() and owner:IsPlayer() and owner.GetPos and id ~= ply:UniqueID() then
 				for key, outfit in pairs(outfits) do
-					if not outfit.wear_filter or table.HasValue(outfit.wear_filter, tostring(ply:UniqueID())) then
+					if not outfit.wear_filter or table.HasValue(outfit.wear_filter, ply:UniqueID()) then
 						pace.SubmitPart(outfit, ply)
 					end
 				end
--- a/lua/pac3/editor/server/wear.lua
+++ b/lua/pac3/editor/server/wear.lua
@@ -408,7 +408,7 @@ function pace.RequestOutfits(ply)
 
 			if owner:IsValid() and owner:IsPlayer() and owner.GetPos and id ~= ply:UniqueID() then
 				for key, outfit in pairs(outfits) do
-					if not outfit.wear_filter or table.HasValue(outfit.wear_filter, tonumber(ply:UniqueID())) then
+					if not outfit.wear_filter or table.HasValue(outfit.wear_filter, tostring(ply:UniqueID())) then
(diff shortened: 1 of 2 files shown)


lua/pac3/editor/server/wear.lua
```diff
@@ -408,7 +408,7 @@ function pace.RequestOutfits(ply)
 
 			if owner:IsValid() and owner:IsPlayer() and owner.GetPos and id ~= ply:UniqueID() then
 				for key, outfit in pairs(outfits) do
-					if not outfit.wear_filter or table.HasValue(outfit.wear_filter, tonumber(ply:UniqueID())) then
+					if not outfit.wear_filter or table.HasValue(outfit.wear_filter, tostring(ply:UniqueID())) then
```
CapsAdmin: UniqueID already returns a string so tostring is redundant

Techbot121: guess the wiki is wrong then? https://wiki.garrysmod.com/page/Player/UniqueID

thegrb93:
Ah that probably fixes all the wear filter issues. Good catch.

roboderpy:
If writeuint with 32 bits overflown from util.CRC (which is used internally by UniquieID), then something very bad happened in case where overflow occured.

thegrb93:
metastruct probably uses its own overrided uniqueid function

CapsAdmin:
Looking at the wiki page for uid regarding collisions it's probably best to stick with steamids. It's more portable too

roboderpy:
Chances for such collision is *extremely unlikely*, however, one darkrp server owner already hit such collision, but only in database context. Uids are still somewhat reliable for short term player identify, such as wear filter. But still I'll probably change uids to steamid as caps suggested since they don't create much traffic if would be used in wear filter. However, there are also parts of pac3 code which should be changed to use steamids instead of uids. 13:26, 7 октября 2019 г., Elias Hogstvedt <notifications@github.com>:Looking at the wiki page for uid regarding collisions it's probably best to stick with steamids. It's more portable too

—You are receiving this because you commented.Reply to this email directly, view it on GitHub, or mute the thread.
-- Отправлено из мобильного приложения Яндекс.Почты

PR #850, opened 2019-12-09
check if UniqueID function exist

FarukGamer:
I'm not sure how the error gers triggerd, but I guess it's because the player or outfit isn't fully loaded.

> addons/pac3-develop/lua/pac3/core/client/parts/custom_animation.lua:37: attempt to call method 'UniqueID' (a nil value)

This was found while playing on the Gamemode Jazztronauts
The errors occured on the of the cutscenes.
Steps to Reproduce.
1. Get the lastest Dev version of pac3
2. Subscribe to the Addons of Jazztronauts ([Collection](https://steamcommunity.com/workshop/filedetails/?id=1455883814))
3. Launch the map jazz_bar
4. Load a pac with custom animation part
5. Try a cutscene.

### DIFF
--- a/lua/pac3/core/client/parts/custom_animation.lua
+++ b/lua/pac3/core/client/parts/custom_animation.lua
@@ -34,7 +34,7 @@ function PART:GetNiceName()
 end
 
 function PART:GetAnimID()
-	return "pac_anim_" .. (self:GetPlayerOwner():IsValid() and self:GetPlayerOwner():UniqueID() or "") .. self:GetUniqueID()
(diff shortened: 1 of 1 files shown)


roboderpy:
If UniqueID is gone inside cutscenes, Jazztronauts is doing something very wrong

roboderpy:
gmod_11-24-26-1641284
This is indeed an issue on Jazztronauts side - it create a "fake" player entity using ClientsideModel() and apply stuff to it, which cause extreme variety of issues.
gmod_11-26-08-2eb8273
gmod_11-26-33-e188eeb

PR #865, opened 2020-03-01
boneanimlib a nil value

Heyter:


thegrb93:
What's boneanimlib?

DBotThePony:
boneanimlib is (still?) used for Lua based animations

Heyter:
I don't know what it is, just fixed the error.

thegrb93:
I'm guessing this option is there to counteract some 3rd party library. This change looks good.

thegrb93:
I think the option text should be changed to 'Stop all boneanimlib animations', but idk if that's better or worse.

Techbot121:
Just for future reference:
[boneanim lib was replaced by "animations.lua" in libraries.](https://github.com/CapsAdmin/pac3/commit/237cacaad5d132663cfa0c76631b84addbf960d9) 
[Originally it was made by jetboom](https://github.com/JetBoom/animationsapi) and I guess caps modified it and changed the name.

thegrb93:
Thanks. I've corrected it.

Issue #888, opened 2020-05-16
Non-physical ticked projectiles only renders one at a time


TheReeb:
![projectile stuff](https://user-images.githubusercontent.com/65454681/82121804-0f89df00-9790-11ea-94c2-889bdffd76af.png)
As the title says, unless you tick the "physical" setting on a projectile, it will only render 1 at a time. I have provided a picture above and an example outfit below to demonstrate (FYI: red ball is physical, blue ball is non-physical)). 

It is worth noting that while the blue ball only renders 1 at a time, it STILL exists until it expires.

Example outfit: https://www.dropbox.com/s/ibhj8vp0luqyxel/projectile%20outfit%20problem.txt?dl=0

thegrb93:
Looks like the problem is pac.HookEntityRender uses ent:EntIndex() and clientside entities all have the same index of -1.

thegrb93:
I'll test the fix out in a little while

thegrb93:
Fixed in develop.

PR #890, opened 2020-05-16
Made rendertables use ent instead of entindex. Maybe fixes: #888

thegrb93:


thegrb93:
It works. Didn't notice any issues with other pacs as well.

Issue #859, opened 2020-02-10
Full network update invalidate (and remove most of) parts and require their re-wear


devSera:
If your client full updates any Parts that have been created by a dupe disappear and do not return. In my own testing I have found that setting part.dupe_remove to nil fixes it with no issues. Not sure what the purpose of dupe_remove is.

https://github.com/CapsAdmin/pac3/blob/97ab99e9e8f5f16063ee5480ee33d21970822b8c/lua/pac3/core/client/part_pool.lua#L464-L484

thegrb93:
Looks like its there so the parts get removed when the player disconnects from the server.

devSera:
Perhaps there could be a timer to see if the player has truly left? As it is, if the server freezes for a bit, all of your duped props lose their pac objects.

Python1320:
it can take over a minute before you know from player entity alone if one has disappeared. Better just network when it's okay to purge an item?

thegrb93:
A timer.simple(0 is enough to get around the EntityRemoved issue. Starfall uses that

DBotThePony:
Full network update invalidate a lot of things, not only duped entities are affected but also players with outfits out of PVS sight.

This is a known issue

thegrb93:
You can see if https://github.com/CapsAdmin/pac3/pull/893 fixes it now

thegrb93:
@devSera

devSera:
It does seem to fix the issue

PR #892, opened 2020-05-18
Maybe fixes: #859

thegrb93:


thegrb93:
Fucks sake, can the default branch be develop?

CapsAdmin:
done

Issue #868, opened 2020-03-22
MDL zips should check for existing models in the game and skip not found ones


Cynosphere:
> [PAC3] the model want to include models/pac3_cache/1315907695/f_anm.mdl but it doesn't exist
> [PAC3] the model want to include models/pac3_cache/1315907695/custom_animation_female.mdl but it doesn't exist

f_anm.mdl is a stock GMod model used by all player models
custom_animation_female.mdl was from an older version of Custom Taunt but is entirely optional

Whether or not a model should be optional is going to be hard to determine is the issue.

[Model used](https://steamcommunity.com/sharedfiles/filedetails/?id=1844633740)

WTTG2Noir:
bc those are not valid paths like before compiling the f_anm.mdl etc needs to be in the given folder of the $cdmaterials "<path>"

Cynosphere:
> f_anm.mdl is a stock GMod model used by all [female] player models

thegrb93:
So I guess the behavior should be modified so if it can't find the model, it checks if it exists in the game?

CapsAdmin:
Sounds reasonable to me. Not sure how to generalize this though. 

How does the code know where to find f_anm.mdl if it's not included? Should we use a lookup table or does mdl support some sort of fallback mechanism? Lookup table for models like f_anm is probably fine for now.

thegrb93:
I'm current experimenting with it

Yagira:
You know what would be better, being able to include additional mdl files in your zip so you could actually import models that have custom animation sets.

thegrb93:
That would be good in addition to this.

thegrb93:
I got the other animation errors fixed, but still get `[PAC3] the model want to include models/pac3_cache/1370833833/custom_animation_female.mdl but it doesn't exist
`

thegrb93:
You can try the fix in that pull request and let me know how it goes.

thegrb93:
@Yagira actually I think it already supports that.

thegrb93:
I'm not sure how it selects which model to be the primary model though..

Yagira:
Yeah well, the thing is, it doesn't really work. Maybe its specific to the model I tried but the animations are completely messed up. Also including the animation .mdl seems to break the materials? There's also a huge fps drops when using it as my entity.
You can try it yourself.
https://dl.dropboxusercontent.com/s/4de40jrynqdl9yf/pac%20test.zip - animation .mdl included
[...]

PR #898, opened 2020-05-26
Fallback to ingame model if zip doesn't contain it

thegrb93:
Possibly Fixes: #868


thegrb93:
I think this needs harder security

Issue #899, opened 2020-05-26
Zip model system $includemodel corrupting a model's material names


thegrb93:
https://dl.dropboxusercontent.com/s/4de40jrynqdl9yf/pac%20test.zip - animation .mdl included
https://dl.dropboxusercontent.com/s/pv98vfcs0xvfpbe/pac%20test2.zip - no animations

Decompiled smd showing the corrupted material names
[model_Body.txt](https://github.com/CapsAdmin/pac3/files/4682200/model_Body.txt)

thegrb93:
Here's the generated .gma. 
[3265074193.zip](https://github.com/CapsAdmin/pac3/files/4682214/3265074193.zip)

thegrb93:
Looks like the $includemodel worked fine, but it broke the materials.

thegrb93:
The issue is offsets are not recomputed properly. .mdl is just data plopped on the end of one another and linked using pointers. The best solution is to append new data to the end and change the pointers to the new data.

PR #903, opened 2020-05-28
Started mdl corruption fix

thegrb93:
Fixes: #899


thegrb93:
@Yagira well the model works now, but it still lags to shit. Idk why.

Yagira:
@thegrb93 apparently its caused by the lack of the .ani file, removing it from the addon causes the same effect as in pac (lag + messy animations), however even after including the .ani file it still doesn't work in pac.

https://dl.dropboxusercontent.com/s/3qoeuan58a750wi/pac%20test3.zip

PR #910, opened 2020-05-31
Fix parts not being unhidden

thegrb93:


thegrb93:
Fixed the parts' ownership issues so far, but now need to figure out why parts do not unhide.

PR #913, opened 2020-06-01
this is about the custom models importing and mostly the material part

RubYNameIess:
ok so you know that you can import custom models, objs and mdls (now talking about objs), well you can import custom materials for them as well, BUT some models have multiple materials that can be applied to them and changes different parts, if possible make it so you can make a folder (lets call it as "materials") and into that folder you can put all the materials for the model and putting that folder to dropbox you can copy it's link and put it to the "material" part in the model in gmod so it applies all of the material to the model (i need this for destiny island from kingdom hearts) if this is possible please make it a thing, also if possible make a thing where you can apply .ani files (contains animations) like i have one that basicly has fortnite and mmd dances and i would like to apply it to my model so if we had an entry that allows player to apply .ani files to a model that would be nice, but if not possible it's fine all i need is the material thingy

### DIFF
--- a/lua/pac3/core/client/base_part.lua
+++ b/lua/pac3/core/client/base_part.lua
@@ -14,16 +14,16 @@ local function SETUP_CACHE_FUNC(tbl, func_name)
 	local old_func = tbl[func_name]
 
 	local cached_key = "cached_" .. func_name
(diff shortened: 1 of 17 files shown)


thegrb93:
obj's can only have one material applied to them. you can split your multi-material obj into seperate objs.

PR #916, opened 2020-06-04
Fix text getting cut off

FarukGamer:
This should fix text which got cutoff.

### DIFF
--- a/lua/pac3/core/client/parts/text.lua
+++ b/lua/pac3/core/client/parts/text.lua
@@ -3,6 +3,7 @@ local cam_Start3D2D = cam.Start3D2D
 local EyePos = EyePos
 local EyeAngles = EyeAngles
 local draw_SimpleTextOutlined = draw.SimpleTextOutlined
+local surface_DisableClipping = surface.DisableClipping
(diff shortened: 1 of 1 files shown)


thegrb93:
I remember surface.DisableClipping( false ) causing issues in starfall. I think true is the default.

DBotThePony:
text should be really long for this to be an issue

CapsAdmin:
surface.DisableClipping is for clipping when text and other things goes outside of the viewport. It makes sense to disable this when doing things in 3d.  Maybe at a higher level though, and not per text part?

It's a new function and I can't recall ever having issues with this since after it was introduced. Maybe some optimization update was the reason.

thegrb93:
Here's the related starfall issue @DBotThePony @CapsAdmin @FarukGamer https://github.com/thegrb93/StarfallEx/issues/747

thegrb93:
This was the fix. https://github.com/thegrb93/StarfallEx/commit/641e1c96279ac04cf3be9aafc7e17e2e5ca856aa

I think the issue is other addons set it to true and then back to false without realizing it's supposed to stay true.

PR #972, opened 2021-01-24
Disabled the language ID indicator

zxz41:
It was constantly getting in the way, hiding up to 2-3 characters. I'm supriced nobody did this long ago.


thegrb93:
What does the language identifier look like? Not exactly sure what this is changing.

thegrb93:
Oh, is this for non-US keyboards I guess?

Yagira:
Yeah, happens if change the keyboard to a different language. 
image

thegrb93 (approved):
I think the changes look good.

Issue #976, opened 2021-01-27
Idea: Command wheel for pac_events


TW1STaL1CKY:
Creating an issue so the idea isn't lost in the Discord

A concommand (+pac_wheel/-pac_wheel ?) that displays a command wheel HUD showing you "command" events you could trigger, in the same way you can do stuff using the Expressions wheel from VRChat.

The command wheel would populate itself by scanning through the localplayer's PAC for "command" events that meet criteria.
Each command option needs a way to either fire one-off (runs "pac_event thing") or toggle (runs "pac_event thing 2"). Maybe via single-click and double-click?

Two new settings can be added to command events for managing the command wheel:
- **show on command wheel** - true/false - enables the command to appear on the wheel
- **command wheel folder** - string - places commands inside a folder within the wheel for organisation

Hook could be added for developers:
- **PreShowCommandWheel** - runs to check if the wheel should render (eg. server owner might want to disable the use of the wheel, or someone might want to make their own version of the wheel)

thegrb93:
I have used this for my starfall emote chip. Adding to pac would be a great addition.

CapsAdmin:
> I have used this for my starfall emote chip. Adding to pac would be a great addition.

maybe you can add it to pac?

thegrb93:
> > I have used this for my starfall emote chip. Adding to pac would be a great addition.
> 
> maybe you can add it to pac?

I can do graphics and interface if you can add the new event type and table of them that the selector can read.

PR #980, opened 2021-01-30
Custom event selector

thegrb93:
Fixes: #976


thegrb93:
@CapsAdmin is sgm/playercircle a gmod material? If not we need to add that resource. Otherwise, just delete the .png and it should be ready to merge.

CapsAdmin:
Somehow I didn't see the png committed, and in game it was originally an error so I assumed you left it out.

thegrb93:
Tested and looks ready to merge

PR #996, opened 2021-02-10
Port over the new StringStream and net streaming functionality from starfall

CapsAdmin:
This is not 100% yet, there's currently an infinite loop in mdlzip functionality.

### DIFF
--- a/lua/pac3/core/shared/util.lua
+++ b/lua/pac3/core/shared/util.lua
@@ -377,7 +377,7 @@ function pac.DownloadMDL(url, callback, onfail, ply)
 							file.Write(data.file_name..".debug.old.dat", data.buffer)
 						end
 
-						local f = pac.StringStream(data.buffer, 1, "little")
+						local f = pac.StringStream(data.buffer)
 						local id = f:read(4)
 						local version = f:readUInt32()
 						local checksum = f:readUInt32()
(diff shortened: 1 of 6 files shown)


lua/pac3/core/shared/util.lua
```diff
@@ -647,37 +647,43 @@ function pac.DownloadMDL(url, callback, onfail, ply)
...
 
-						-- Add nulls to align to 4 bytes
-						while (#f.buffer)%4 ~= 0 do
-							f.buffer[#f.buffer+1] = 0
+
+						do
+							error("NYI, code below infinite loops")
+
+							-- Add nulls to align to 4 bytes
+							while (#f)%4 ~= 0 do
```
CapsAdmin, line 670: have a look here @thegrb93

thegrb93:
I'll correct it. The new version has some differences. Also Seek and Tell are 1 first index like lua instead of 0 first index.

thegrb93:
Works with my outfit. 👍

thegrb93:
I think that's everything now. Might want @Yagira to also test that it still works.

Yagira:
> Might want @Yagira to also test that it still works.

All my models seem to work fine.

CapsAdmin:
I think that's good enough for the develop branch.

thegrb93:
@CapsAdmin One thing I just remembered. The reason to make net_stream global is so that it is shared between starfall/advdupe2/pac3. That way two won't be running concurrently and potentially overflow the net buffer.

thegrb93:
It might be a good idea to switch it to be global again.

PR #1012, opened 2021-02-21
add ability to set model to playermodel

2dve:


CapsAdmin:
Can you explain the need for this? I think I would do it some other way if I understand the need correctly.

You either want the model to reflect the local player's model, meaning if I'm kleiner the model will be kleiner, but if someone is alyx the model will be alyx for them, but kleiner for me.

Or you want the model to reflect the outfit's player model.

2dve:
Thank you, this has been fixed. The feature is to set the model to the outfit wearer's playermodel. I used this to achieve the effect of replacing the right arm, which would have been impossible otherwise due to the fact that weapons scale with the right hand bone.

2dve:
basically instead of having to manually use events and custom weapon models to make the weapons appear, i just have it set up as pictured with the playermodel being invisible and the weapon still showing
image
image

2dve:
This feature can also make it easier for server developers to add items that require the player model to be used, as there is no way currently for PAC to automatically change a model to the current player model.

CapsAdmin:
I think instead of doing this, I would make it so you can select "player_model" as a model in the model field.

lua/pac3/core/client/parts/model.lua
```diff
@@ -405,6 +406,16 @@ function PART:DrawModel(ent, pos, ang)
 	if self.loading and not self.obj_mesh then return end
 
 
+	if self.SetModelToPlayer then
```
thegrb93, line 409: Doing this every frame is not good for performance

PR #1022, opened 2021-03-19
replace some type with is*

FarukGamer:
C `is*` funcs are faster then `type`, since a string doesn't have to be pushed to Lua, and the code doesn't have to return to C for the string comparison.

### DIFF
--- a/lua/pac3/core/client/part_pool.lua
+++ b/lua/pac3/core/client/part_pool.lua
@@ -517,7 +517,7 @@ function pac.UpdatePartsWithMetatable(META, name)
 	for _, part in pairs(all_parts) do
 		if part.ClassName == name then
 			for k, v in pairs(META) do
-				if type(v) == "function" then
+				if isfunction(v) then
 					part[k] = v
 				end
 			end
--- a/lua/pac3/core/client/parts/beam.lua
+++ b/lua/pac3/core/client/parts/beam.lua
@@ -183,7 +183,7 @@ function PART:SetMaterial(var)
 	self.Material = var
 
 	if not pac.Handleurltex(self, var) then
-		if type(var) == "string" then
+		if isstring(var) then
 			self.Materialm = pac.Material(var, self)
 			self:FixMaterial()
 			self:CallEvent("material_changed")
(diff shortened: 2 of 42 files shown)


CapsAdmin:
Do you see a performance benefit from doing this?

CapsAdmin:
I prefer using type as it's a standard lua function. But I can get onboard with using isangle, isvector, etc for non standard lua types.

The exception I make is for performance, but I shouldn't be comparing types in performance sensitive code anyway. Looking at the code that's changed, I mostly see changes in code that is not being called often.

thegrb93:
I think jit would normally optimize them (if garry hadn't botched the type function) so these are the optimal way to do it in gmod.

thegrb93:
That said, that may not always be the case if https://github.com/Facepunch/garrysmod-issues/issues/4100 ever gets resolved.

thegrb93:
We could git revert if things change though.

thegrb93:
Just reviewed the changes and they look one to one so it's just a matter of whether we want the change.

FarukGamer:
@CapsAdmin Just take a look here. Evetho the link refers to a ulib pull request. Somebody did benchmark it.

https://github.com/TeamUlysses/ulib/pull/58#issuecomment-518730448

CapsAdmin:
I believe you when you say the function is faster, but when it comes to performance in PAC3 you should be comparing performance for when an outfit is being rendered.

If you want an even faster type and is* function, see 

https://github.com/PAC3-Server/notagain/blob/master/lua/notagain/optimizations/preinit/luaify.lua#L6-L76

You can apply this globally and all addons will benefit from it without having to resort to writing gmod specific lua.

FarukGamer:
It ment to be just some micro optimisation. That it will improve the **runtime performance** . But do not stick to my words. You may know better, I just wanted to submit a possible improvement.

thegrb93:
No sense in wasting perfectly good work. Just poke Caps until he merges it.

thegrb93:
> https://github.com/PAC3-Server/notagain/blob/master/lua/notagain/optimizations/preinit/luaify.lua#L6-L76

This is disgusting. I'm going to try it out

FarukGamer:
@thegrb93 If your tests for that brings nothing good, then will open it back.

thegrb93:
I think it ought to be merged either way, but ya I'll see if it helps starfall at all

thegrb93:
Seems jit.off() still faster than that optimization

PR #1025, opened 2021-04-04
Prettyprint code base

CapsAdmin:
Prettyprint everything with NattLua, do you guys see any glaring issues?

The config at the moment is:
https://github.com/CapsAdmin/pac3/blob/linter/lint.lua#L38-L65


thegrb93:
Looks good for the official gmod codebase but personally more a fan of more condensed code.

PR #1026, opened 2021-04-04
prevent disapearing parts affter death on servers where is no weapon on spawn

owltaver:
So when you spawn with no weapon, you outfit gonna be hidden, becouse it loads only from OnEntityCreated hook from your weapon(s) via GetOwner. There is a fix for it. Also is there reason why hide outfit cvar was removed?

### DIFF
--- a/lua/pac3/core/client/part_pool.lua
+++ b/lua/pac3/core/client/part_pool.lua
@@ -663,7 +663,6 @@ do -- drawing
 		local dummyv = Vector(0.577350,0.577350,0.577350)
 		local fovoverride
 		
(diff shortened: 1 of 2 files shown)


CapsAdmin:
The cvar was removed because by default the outfit would render on the invisible player rather than the ragdoll. Now it renders on the ragdoll so I thought I could remove it. I don't see the need for the cvar anymore.

You can change this behavior yourself on your outfit with the player config part.

owltaver:
> The cvar was removed because by default the outfit would render on the invisible player rather than the ragdoll. Now it renders on the ragdoll so I thought I could remove it. I don't see the need for the cvar anymore.
> 
> You can change this behavior yourself on your outfit with the player config part.

Understandable, thanks for the answer.

PR #1069, opened 2021-05-11
Let some special bones be animated

luttje:
Hey hey,

**Summary: With this PR I show how few modifications are needed to allow those special (fake) bones to animate**

So I ran into a bit of a struggle making a skateboard outfit. I wanted to create a Custom Animation to animate tricks, though binding to existing bones caused the skateboard to move with the players' body and breath.
Having the skateboard completely separate and having two Custom Animations was a hassle to synchronize, since board movement and feet/leg movement are closely related in skateboarding.

I wanted to use a "fake" bone which is not actually bound to the model. More like the special bones (skirt2, pos_ang, etc.) defined in `core/client/bones.lua`, but those were impossible to animate.

I've written up a prototype which allows select special bones to be animated. It is just an initial prototype/concept, please do comment on what you think of it and how you'd like to see this implemented. A point of improvement I think would be valuable would be the ability to add "Special Bones" through hooks, or perhaps even through the pac editor.

I hope I've made clear why this can be useful, here's some additional screenshots I took:

bind to a special bone (was already possible)
animate the special bone (new)
have the model animate completely independently from the model

### DIFF
--- a/lua/pac3/core/client/bones.lua
+++ b/lua/pac3/core/client/bones.lua
@@ -7,6 +7,7 @@ local util_QuickTrace = util.QuickTrace
 local pac = pac
 local pac_isCameraAllowed = pac.CreateClientConVarFast("pac_enable_camera_as_bone", "1", true, "boolean")
 
+pac.SpecialBoneNamePrefix = "$special_"
 pac.BoneNameReplacements =
 {
 	{"Anim_Attachment", "attach"},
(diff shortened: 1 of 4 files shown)


lua/pac3/core/client/bones.lua
```diff
@@ -366,6 +406,31 @@ do -- bone manipulation for boneanimlib
...
+				ent.pac_special_bone_anims[boneID] = nil
+			end
+		end
+	end
+
+	function pac.ManipulateSpecialBone(ent, boneID, pos, ang)
+		if(ent.pac_special_bone_anims and ent.pac_special_bone_anims[boneID])then
+			local matrix = ent.pac_special_bone_anims[boneID]
+			pos = pos + matrix:GetTranslation()
+			ang = ang + matrix:GetAngles()
```
thegrb93, line 426: This should be somethink like:
```lua
result = matrix1 * matrix2
pos = result:GetTranslation()
ang = result:GetAngles()
```
Adding euler angles is almost never good

thegrb93 (changes requested):

lua/pac3/core/client/bones.lua
luttje: I didn't know that, thanks! Is it because of gimbal lock?

I've just pushed a commit with what I think is a proper way of implementing what you described.

thegrb93: Looks good. Yeah gimble lock and also because adding euler angle values won't accurately represent a rotation plus another rotation. It only does if only one of the components of each are nonzero.

Issue #1066, opened 2021-05-10
ForceObjUrl became false when wearing (legaсy convertation).


owltaver:
After convert legacy to new obj part, obj loads normaly, but not when you wear it. For some reason ForceObj became false when wearing and obj(s) start loading like mdl(s) causing console mdl dl error spam.
Outfit used for testing: https://pastebin.com/r5qiV0gv (convert via tools>wear)

Techbot121:
seems to work now

PR #1072, opened 2021-05-28
better typecheck

owltaver:
Just noticed that forceobj does not work on SetModel in most cases, so this is not only [convertation issue](https://github.com/CapsAdmin/pac3/issues/1066). Guess because meta vars is not full loaded at that moment. Anyway there is a some better way to do typechecks and make it automated.

### DIFF
--- a/lua/pac3/core/client/parts/model.lua
+++ b/lua/pac3/core/client/parts/model.lua
@@ -32,7 +32,6 @@ BUILDER:StartStorableVars()
 		:PropertyOrder("Hide")
 		:PropertyOrder("ParentName")
 		:GetSet("Model", "", {editor_panel = "model"})
-		:GetSet("ForceObjUrl", false)
 
 	:SetPropertyGroup("orientation")
 		:GetSet("Size", 1, {editor_sensitivity = 0.25})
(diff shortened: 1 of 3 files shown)


CapsAdmin:
Looks good. The only thing I'm worried about is potential breaking so I'd like to test some cases myself first.

CapsAdmin:
When using onedrive the HEAD response is 
```Cache-Control	=	no-cache, no-store
Content-Length	=	0
Date	=	Mon, 21 Jun 2021 22:26:55 GMT
Expires	=	-1
Pragma	=	no-cache
Set-Cookie	=	E=P:Qv00rQM12Yg=:5geWaawr81fpzUZe4cPKaTdnYLnVpspnAYZdHPBW5vM=:F; domain=.live.com; path=/
```
So Content-Type is missing. I think there should be a fallback of some sort

thegrb93:
What was the response code there? 204?

CapsAdmin:
Right, it's actually 404. Probably because onedrive doesn't support HEAD requests.

I'd post the url but there's a scary authKey query param in it and it's not my url. 😅

PR #1096, opened 2021-10-05
Return result of pac.CallHook and pace.CallHook

brandonsturgeon:
`pac.CallHook` and `pace.CallHook` (serverside) don't return anything, but they're sometimes used as if they did.

I can't tell if this is intended or not. The clientside versions of these functions return the result.

This is a problem because of lines like this:
https://github.com/CapsAdmin/pac3/blob/master/lua/pac3/editor/server/wear.lua#L412-L414

Returning `false` in this hook serverside doesn't do anything because `pac.CallHook` doesn't return anything.

If this is undesirable for some reason, we could also refactor some code to make sure nothing expects these functions to return anything.

### DIFF
--- a/lua/autorun/pac_version.lua
+++ b/lua/autorun/pac_version.lua
@@ -1,39 +1,5 @@
 AddCSLuaFile()
 
-local hash = include("pac3/libraries/hash_version.lua")
-
-local pac3
-local core
-local editor
-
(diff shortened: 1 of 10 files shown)


thegrb93 (approved):

DBotThePony:
merged

Issue #1089, opened 2021-08-21
[develop] using an mdl zip as entity causes freezes/crashes 


Yagira:
Seems like it started happening after some of the recent changes. 
Model link: https://wiki.pac3.info/_media/tutorial/mdl/animegirl.zip

thegrb93:
Reverted closing PR so reopening.

Issue #1097, opened 2021-10-06
entity mutator errors all the time


Earu:
On startup, when someone wears a pac, almost all the time even
```
lua/pac3/core/shared/entity_mutator.lua:78: player is invalidlua/pac3/core/shared/entity_mutator.lua:78: player is invalid
stack traceback:
	lua/pac3/core/shared/entity_mutator.lua:70: in function <lua/pac3/core/shared/entity_mutator.lua:69>
	[C]: in function 'assert'
	lua/pac3/core/shared/entity_mutator.lua:78: in function 'MutateEntity'
	lua/pac3/core/shared/entity_mutator.lua:364: in function <lua/pac3/core/shared/entity_mutator.lua:360>
	[C]: in function 'xpcall'
	lua/pac3/core/shared/entity_mutator.lua:360: in function <lua/pac3/core/shared/entity_mutator.lua:353>
	[C]: in function 'xpcall'
	lua/includes/extensions/net.lua:86: in function <lua/includes/extensions/net.lua:56>
lua/pac3/core/shared/entity_mutator.lua:78: player is invalidlua/pac3/core/shared/entity_mutator.lua:78: player is invalid
stack traceback:
	lua/pac3/core/shared/entity_mutator.lua:70: in function <lua/pac3/core/shared/entity_mutator.lua:69>
	[C]: in function 'assert'
	lua/pac3/core/shared/entity_mutator.lua:78: in function 'MutateEntity'
	lua/pac3/core/shared/entity_mutator.lua:364: in function <lua/pac3/core/shared/entity_mutator.lua:360>
	[C]: in function 'xpcall'
	lua/pac3/core/shared/entity_mutator.lua:360: in function <lua/pac3/core/shared/entity_mutator.lua:353>
	[C]: in function 'xpcall'
	lua/includes/extensions/net.lua:86: in function <lua/includes/extensions/net.lua:56>
lua/pac3/core/shared/entity_mutator.lua:78: player is invalidlua/pac3/core/shared/entity_mutator.lua:78: player is invalid
stack traceback:
	lua/pac3/core/shared/entity_mutator.lua:70: in function <lua/pac3/core/shared/entity_mutator.lua:69>
	[C]: in function 'assert'
	lua/pac3/core/shared/entity_mutator.lua:78: in function 'MutateEntity'
	lua/pac3/core/shared/entity_mutator.lua:364: in function <lua/pac3/core/shared/entity_mutator.lua:360>
[...]

PR #1100, opened 2021-10-16
Entity fix

### LINKED ISSUE #1098: pac_restart causes entity desync
If you use `pac_restart` while SOMEONE ELSE is using the entity part they will get stuck in place or start lagging/teleporting/whatever on your end. This is not a new problem, it's been a thing for a while. It persists until you restart/rejoin the server. 

Video: https://drive.google.com/file/d/1k2OaKZE5T8OLoc3hHWKyBJW9bUPylgER/view?usp=sharing

If you use `pac_restart` while YOU are using the entity part you will end up in that weird state with broken animations or something. Also persists until restart (?).

Video: https://drive.google.com/file/d/1LzYuXeZ3EyRMkdGJhMZKSMkAeWK1X_sI/view?usp=sharing

As a bonus this breaks the model_name event even since apparently `:GetModel()` returns the wrong model (???) after a restart.

Model used: https://wiki.pac3.info/_media/tutorial/mdl/animegirl.zip

Yagira:
Should fix #1098 and #1089 as well as random error models and broken ragdolls. Hopefully it doesn't break anything.

### DIFF
--- a/lua/pac3/editor/client/menu_bar.lua
+++ b/lua/pac3/editor/client/menu_bar.lua
@@ -33,6 +33,10 @@ local function populate_pac(menu)
 	end
 
 	do
+		menu:AddOption(L"request", function() RunConsoleCommand("pac_request_outfits") pac.Message('Requesting outfits.') end):SetImage(pace.MiscIcons.replace)
+	end
+
+	do
 		local menu, icon = menu:AddSubMenu(L"clear", function() end)
 		icon:SetImage(pace.MiscIcons.clear)
 		menu.GetDeleteSelf = function() return false end
--- a/lua/pac3/editor/server/wear_filter.lua
+++ b/lua/pac3/editor/server/wear_filter.lua
@@ -4,9 +4,11 @@ util.AddNetworkString('pac_update_playerfilter')
 
 local function find_outfits(ply)
 	for id, outfits in pairs(pace.Parts) do
-		local owner = player.GetByUniqueID(id) or NULL
-		if owner == ply then
-			return outfits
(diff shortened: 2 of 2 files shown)


Techbot121:
I'm currently testing this on metastruct, works fine so far, only downside of this you have to clear the pac3_cache of both **local and server**

thegrb93:
What's wrong if you don't clear the cache?

Yagira:
We thought that maybe it was cache related because some people kept seeing errors but that doesn't really make sense because the ids changed anyway.

I think that the actual problem is the way pac works. The server got restarted and some people joined to try the new version, they wore their outfits etc. Then some more people joined and that is where the problem starts, because you need to be on the server when someone wears the outfit in order to see it (this affects outfits in general not just the entity). If you don't see the outfit you never download the model so you just end up with an error. To fix it you have to either wait for people to re-wear or just ask them to do so which is kinda silly.
 
Maybe outfits should auto wear every X minutes or players should somehow receive them when joining the game / be able to request them on their own.

Techbot121:
what you're looking for is here 
https://github.com/CapsAdmin/pac3/blob/develop/lua/pac3/core/shared/entity_mutator.lua#L307-L317

This should send the players the entity mutators when they first move, but I guess it doesn't?
I'm just speculating since I didn't actually try it ingame yet.

thegrb93:
Seems to have caused https://github.com/CapsAdmin/pac3/issues/1097 . Could you rework your fixes and resubmit? I can provide a reproduction outfit if needed.

PR #1112, opened 2021-12-01
Entity/model fixes once again

Yagira:


Yagira:
Should fix #1097 and #1098 
@thegrb93 I tested this with your outfit and it seems fine but would be nice if you could check it yourself and see if it works

thegrb93:
Looks like that fixed it. Thanks

PR #1113, opened 2021-12-02
Proxy fixes

Yagira:


thegrb93 (approved):

lua/pac3/core/client/parts/proxy.lua
```diff
@@ -192,6 +192,30 @@ end
 
 PART.Inputs = {}
 
+PART.Inputs.property = function(self, property_name, field)
```
thegrb93, line 195: I'm a bit concerned about the performance drop from this. Can you provide a test outfit where I can investigate what calls this functions?

Yagira: [property_proxy.txt](https://github.com/CapsAdmin/pac3/files/7646693/property_proxy.txt)

thegrb93:
Ok, I think it looks good. Thanks

PR #1115, opened 2021-12-02
Fixed animation part

Jacbo1:
- Rate setting now multiplies playback rate so a Rate of 1 is now actually 1x playback speed
- Fixed divide by 0 when Min and Max settings are the same number (doing this now explicitly sets the progress of the animation)
- Setting Rate to 0 now pauses animation like one would expect instead of also setting the current frame to 0
- Removed rate limit that was sometimes so low that animations could not be played at the original speed
- Made self.frame wrap from 0 to 1 so extremely high rates will never become so high they experience floating point precision loss
- Removed some unnecessary code that never ran
- Replaced some code with more optimized alternatives

Everything has been tested by me

### DIFF
--- a/lua/pac3/core/client/parts/animation.lua
+++ b/lua/pac3/core/client/parts/animation.lua
@@ -244,54 +244,50 @@ end
 function PART:UpdateAnimation(ent)
 	if not self.random_seqname then return end
 
-	local seq = ent:LookupSequence(self.random_seqname) or 0
-
-	local duration = 0
+	local seq, duration = ent:LookupSequence(self.random_seqname)
+	
 	local count = ent:GetSequenceCount() or 0
-	if seq >= 0 and seq < count and count > 0 then
-		duration = ent:SequenceDuration(seq)
-	else
+	if seq < 0 or seq >= count then
 		-- It's an invalid sequence. Don't bother
 		return
 	end
-
-	local rate = math.min(self.Rate * duration, 1)
-
-	if seq ~= -1 then
-
-		if rate == 0 then
-			ent:SetCycle(self.Offset % 1)
-			return
-		end
-	else
-		seq = tonumber(self.random_seqname) or -1
-
-		if seq ~= -1 then
-			if rate == 0 then
-				ent:SetCycle(self.Offset % 1)
-				return
-			end
-		else
-			return
-		end
-	end
-
-	rate = rate / math.abs(self.Min - self.Max)
-	rate = rate * FrameTime()
-
+	
 	local min = self.Min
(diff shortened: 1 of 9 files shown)


Jacbo1:
I just found out that pull requests are usually made to the develop branch so I'll redo this for that branch.

thegrb93:
Ah yeah. Just do `git rebase develop`

thegrb93:
You can use the edit button in the top right to change the target branch next time instead of closing it.

lua/pac3/core/client/parts/animation.lua
```diff
@@ -244,59 +244,55 @@ end
...
-	if seq >= 0 and seq < count and count > 0 then
-		duration = ent:SequenceDuration(seq)
-	else
+	if seq < 0 or seq >= count then
 		-- It's an invalid sequence. Don't bother
 		return
 	end
+	
+	local min = self.Min
+	local max = self.Max
```
thegrb93, line 256: You could also do `local length = max - min` to optimize away the `max - min` in the later code

Jacbo1:
Well this doesn't directly port to the develop branch

lua/pac3/core/client/parts/animation.lua
```diff
@@ -244,59 +244,55 @@ end
...
+		local cycle = min + math.abs(math.Round(self.frame + self.Offset * 2) - self.frame - self.Offset * 2) * 2 * (max - min)
 
 		if pac.IsNumberValid(cycle) then
 			ent:SetCycle(not self.InvertFrames and cycle or (1 - cycle))
 		end
 	else
-		self.frame = self.frame + rate
-		local cycle = min + ((self.frame + self.Offset) * 0.5) % 1 * (max - min)
+		self.frame = (self.frame + rate) % 1
+		local cycle = min + (self.frame + self.Offset * 2) % 1 * (max - min)
```
thegrb93, line 290: What's the *2 do here?

thegrb93 (changes requested):

lua/pac3/core/client/parts/animation.lua
Jacbo1: I kinda trialed and errored until everything worked right. It's all been tested (on the master branch version)

Jacbo1: It could actually be a mistake from when I was testing with min = 0.5

Jacbo1:
I finished the develop branch version and also fixed something I didn't catch before so I'll close this and make a new one since this is also pulling from a different branch I made.

PR #1117, opened 2021-12-02
Fixed animation

Jacbo1:
- Rate setting now multiplies playback rate so a Rate of 1 is now actually 1x playback speed
- Fixed divide by 0 when Min and Max settings are the same number (doing this now explicitly sets the progress of the animation)
- Setting Rate to 0 now pauses animation like one would expect instead of also setting the current frame to 0
- Removed rate limit that was sometimes so low that animations could not be played at the original speed
- Made self.frame wrap from 0 to 2 so extremely high rates will never become so high they experience floating point precision loss
- Removed some unnecessary code that never ran
- Replaced some code with more optimized alternatives
- Can now disable looping

Everything has been tested by me

### DIFF
--- a/lua/pac3/core/client/parts/animation.lua
+++ b/lua/pac3/core/client/parts/animation.lua
@@ -264,17 +264,17 @@ function PART:OnUpdateAnimation()
 			return
 		end
 
-		local rate = (duration == 0) and 0 or (self.Rate / duration / math.abs(maxmin) * FrameTime())
+		local rate = (duration == 0) and 0 or (self.Rate / duration / math.abs(maxmin) * FrameTime() * 0.5)
 
 		if self.PingPongLoop then
-			self.frame = (self.frame + rate * 0.25) % 1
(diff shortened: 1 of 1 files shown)


thegrb93:
Do the input parameters behave the same or are people going to need to update their outfits?

Jacbo1:
They should behave the same. But you won't need to use a rate of for example 0.031 to get that one specific animation to play at the original speed

Jacbo1:
That last commit is pretty much ignorable. I just wanted to make the number make sense. It was 2 so the normal loop and ping pong loop would sync, then it was 4 when I fixed disabling looping so I could actually clamp it from 0-2, then I realized I don't need to mod it when not looping so I removed the mod and forgot to change the wrapping back to 0-2.

thegrb93:
Could you test again and make sure it works with the recent animation merge?

Jacbo1:
Yeah it still works. I'm not sure why it wouldn't since GitHub says the last change to animation.lua was 10 months ago.

thegrb93:
The animations.lua library was updated. Not sure if the two were depending on one another or not

Issue #1127, opened 2021-12-09
Shake Part range is global


Logzombie:
This has been a bug for ages, or so I assume. The range of the shake part I imagine is supposed to cap how far or when the shake part's effective area is supposed to drop off(?)

To say the least, even on a range as low as 0.1, the shake is still map-wide. I've been very dissuaded from even attempting to use them as the only way to do so without annoying literally everybody on the server is to cap it with a complex proxy set up. _Ideally we shouldn't need very specific workarounds just to prevent a bug caused by a lack of functionality on a part. This isn't convenient to use in the slightest._ Overall, will be useful to have this fixed since this bug has been there for ages and I haven't spotted an open issue on it. So, thought I would bring light to it once again.

thegrb93:
This is the code. I'm guessing `self:GetDrawPosition()` is returning the local player's position or something in the case the owner isn't valid or something like that.

https://github.com/CapsAdmin/pac3/blob/master/lua/pac3/core/client/parts/shake.lua#L17

thegrb93:
Nvm, looking at the gmod wiki, https://wiki.facepunch.com/gmod/util.ScreenShake , the position and radius don't even do anything on client.

thegrb93:
Yeah, should be easy to make shaking only show for the owner.

PR #1136, opened 2021-12-23
scale shake amplitude with distance

Yagira:
A workaround for #1127 
This pretty much creates a fake shake radius by setting amplitude to 0 if the camera distance is higher than the set radius.
Includes optional falloff setting.

### DIFF
--- a/lua/pac3/core/client/parts/shake.lua
+++ b/lua/pac3/core/client/parts/shake.lua
@@ -20,7 +20,7 @@ function PART:OnShow(from_rendering)
 		local eyedistance = position:Distance(pac.EyePos)
 		local radius = math.Clamp(self.Radius, 0.0001, 500)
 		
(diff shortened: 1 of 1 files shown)


thegrb93:
Ah just make our own shake modifications. Why didn't I think of that. Lgtm

thegrb93:
Also you should use 'local' for local vars

Yagira:
True, forgot about that. 
Also the scaling isn't exactly 'hardcoded', players can just turn off falloff and use a `eye_position_distance()` proxy to create custom scaling function if they want to.

thegrb93:
Looks like the engine does it your way actually https://github.com/ValveSoftware/source-sdk-2013/blob/0d8dceea4310fde5706b3ce1c70609d72a38efdf/sp/src/game/server/util.cpp#L789

thegrb93:
Tested and is working

PR #1143, opened 2021-12-28
Editor fixes

Yagira:
A bunch of small fixes.

### DIFF
--- a/lua/pac3/editor/client/panels/editor.lua
+++ b/lua/pac3/editor/client/panels/editor.lua
@@ -325,17 +325,16 @@ function pace.GainFocus(show_editor)
 		if self.allowclick ~= false then
 			self:MakePopup()
 			pace.Focused = true
-			if not show_editor then
-				timer.Remove("pac_editor_visibility")
 
-				self:SetVisible(true)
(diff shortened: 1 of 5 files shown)


lua/pac3/editor/client/panels/editor.lua
```diff
@@ -345,6 +349,10 @@ function pace.KillFocus(show_editor)
 			self:AlphaTo(0, 0.1, 0)
 			self.exit_button:AlphaTo(0, 0.1, 0)
 			self.zoomframe:AlphaTo(0, 0.1, 0)
+
```
thegrb93, line 352: This should probably have a 0.1 sec delay right? Since the AlphaTo animation would be pointless otherwise.

Yagira: Right, should be better now.

thegrb93: Should probably make it a timer.create and do a timer.remove when the other case happens so turning it off and on again quickly won't hide it

thegrb93 (approved):

PR #1147, opened 2022-01-08
use full material path instead of name when needed

### LINKED ISSUE #1146: MDL downloads does not seek proper textures despite existing on -fastbuild MDLs
- [Model ZIP](https://iastg.s3.ap.cloud-object-storage.appdomain.cloud/stg_1928032180128730.zip).

PAC3 does not know where to look for textures despite existing within the ZIP when the MDL is compiled using -fastbuild.

image

Yagira:
fixes #1146

### DIFF
--- a/lua/pac3/core/shared/util.lua
+++ b/lua/pac3/core/shared/util.lua
@@ -402,12 +402,11 @@ function pac.DownloadMDL(url, callback, onfail, ply)
 										local found = false
 
 										if mat:EndsWith("\\.vmt") or mat:EndsWith("/.vmt") or mat == (".vmt") then goto CONTINUE end
-
 										for i, v in pairs(files) do
 											if v.file_name == mat then
 												found = v.file_path
 												break
-											elseif v.file_path == ("materials/" .. mat) or v.file_path:match(mat) or v.file_name == mat:match(".+/(.+)") then
+											elseif v.file_path == ("materials/" .. mat) or string.find(v.file_path, mat, 1, true) or string.find(mat, v.file_name, 1, true) then
 												v.file_name = mat
 												found = v.file_path
 												break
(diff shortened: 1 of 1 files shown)


thegrb93:
Won't this possible cause an issue if there are two materials of same name but in different folders?

Yagira:
Right, this change should handle it as long as you keep the original folder structure when making the zip. It's probably a very rare case anyway.

thegrb93:
I think the rest of the code looks ok. I'm still nervous about matching the filename despite different directory though.

Yagira:
> Why's the `v.file_path:match(mat)` there? This seems strange to me.

Both this and `v.file_name == mat:match(".+/(.+)")` is for cases where someone changed the folder structure, so if there's more folders in the path or there's no `materials` folder or even no folders at all. If the original path doesn't exist then try to match a part of it or the name. 

It's pretty much an attempt to catch things that could be wrong and fix them.

thegrb93:
Oh I see. Can you change that to `string.find(mat, v.file_name, 1, true)`?

Yagira:
This should be fine. Also here is a model zipped in three different ways that I used for testing if you wanna check it out yourself.

[original](https://h.yag.ovh/f/nbisarin.zip)
[additional folder](https://h.yag.ovh/f/folder.zip)
[no folders](https://h.yag.ovh/f/rush1.zip)

thegrb93:
The second match was fine. Could also be `v.file_name==string.GetFileFromFilename(mat)`

thegrb93:
I'll take a look at your examples in a bit

thegrb93:
Can you explain what causes each of your examples? Where do these cases arise? I looked in each .mdl and they are exactly the same so I'm thinking something with the directory structure needs to be improved.

Yagira:
The examples are just the common ways that people make their mdl .zips. They were just to show that even 'broken' zips will still work in PAC. You can load any of the example links in game and you will end up with a proper .gma no matter which one you use.

thegrb93:
Ah so this is just a workaround for #1146 . I'm still very nervous about this change. For example, you can have a model with different clothes bodygroups with materials of the same name in different folders. It'd be a good idea to be sure this doesn't break those.

Issue #1153, opened 2022-02-10
math.randomseed is set to CurTime, making it 100% predictable


MeteorTheLizard:
> addons\pac3\lua\pac3\libraries\hash_version.lua

Line: 121

Detour:
lua/includes/util.lua

````
--
-- Seed the rand!
--
math.randomseed( os.time() )

local mathrandomseed = math.randomseed

math.randomseed = function(iVar)
	print("Seed changed! " .. iVar)

	print(debug.Trace())

	mathrandomseed(iVar)
end
````

Result:
```
Seed changed! 1808

Trace:
	1: Line 32	"Trace"		lua/includes/extensions/debug.lua
	2: Line 12	"randomseed"		lua/includes/util.lua
	3: Line 112	"LuaPaths"		addons/pac3/lua/pac3/libraries/hash_version.lua
	4: Line 43	"PAC_VERSION"		addons/pac3/lua/autorun/pac_version.lua
	5: Line 65	"nil"		addons/pac3/lua/autorun/pac_version.lua


Seed changed! 1.0151515007019

Trace:
	1: Line 32	"Trace"		lua/includes/extensions/debug.lua
	2: Line 12	"randomseed"		lua/includes/util.lua
	3: Line 121	"LuaPaths"		addons/pac3/lua/pac3/libraries/hash_version.lua
	4: Line 43	"PAC_VERSION"		addons/pac3/lua/autorun/pac_version.lua
	5: Line 65	"nil"		addons/pac3/lua/autorun/pac_version.lua


Seed changed! 2616

Trace:
	1: Line 32	"Trace"		lua/includes/extensions/debug.lua
	2: Line 12	"randomseed"		lua/includes/util.lua
	3: Line 112	"LuaPaths"		addons/pac3/lua/pac3/libraries/hash_version.lua
	4: Line 54	"PAC_VERSION"		addons/pac3/lua/autorun/pac_version.lua
	5: Line 65	"nil"		addons/pac3/lua/autorun/pac_version.lua


Seed changed! 1.0151515007019

Trace:
	1: Line 32	"Trace"		lua/includes/extensions/debug.lua
	2: Line 12	"randomseed"		lua/includes/util.lua
	3: Line 121	"LuaPaths"		addons/pac3/lua/pac3/libraries/hash_version.lua
	4: Line 54	"PAC_VERSION"		addons/pac3/lua/autorun/pac_version.lua
[...]

Issue #1104, opened 2021-10-23
pac_version not working as intended


Techbot121:
While the idea was good, there is a problem if you have a different carriage return (i.e. windows vs linux or downloading pac from the workshop vs github as a zip) it will return a different version and can be a pain if you want to compare versions, even though the code is the same.

thegrb93:
The version should definitely provide a commit id instead of this hash nonsense, or workshop if it is mounted via workshop.

thegrb93:
```lua
function pace.version()
    if steamworks.IsSubscribed( "pac_workshop" ) then
        return "workshop"
    end

    for k,v in pairs(select(2, file.Find( "addons/*", "GAME" ))) do
        if file.Exists("addons/"..v.."/lua/autorun/pac_init.lua", "GAME") then
            local dir = "addons/"..v.."/.git/"
            local head = file.Read(dir.."HEAD", "GAME") -- Where head points to
            if not head then break end

            head = string.match(head, "ref:%s+(%S+)")
            if not head then break end

            local lastCommit = file.Read( dir..head, "GAME")
            if not lastCommit then break end

            return lastCommit
        end
    end

    return "unknown"
end)
```

PR #1156, opened 2022-02-16
Refactor pac3 version. Fixes: #1104 Fixes: #1153

thegrb93:


CapsAdmin (approved):
it was fun (but not functional) while it lasted

CapsAdmin:
Maybe I was too quick to approve. I'm fine with the way it is but have some suggestions.

It would be nice if the workshop version doesn't only return "workshop" but the actual hash that was used to update it.

It would be nice to see a human readable hash as well in the same style as before, but instead curate a list of words to use instead of finding all words in the codebase.

thegrb93:
Could probably print the HEAD branch name as well for git. For workshop, maybe something in the .gma's header can give additional info?

PR #1165, opened 2022-02-27
Micro Optimize object type checking

FarukGamer:
Let's save some CPU cycles by using `is*` functions instead of `type`, by omitting the string comparison of the object type we get returned by `type`.

### DIFF
--- a/lua/autorun/pac_restart.lua
+++ b/lua/autorun/pac_restart.lua
@@ -31,7 +31,7 @@ function _G.pac_ReloadParts()
 					local str = file.Read("addons/" .. dir .. "/lua/" .. new_path, "MOD")
 					if str then
 						local func = CompileString(str, "addons/" .. dir .. "/lua/" .. new_path)
-						if type(func) == "function" then
+						if isfunction(func) then
 							local res = {pcall(func, ...)}
 
 							if res[1] then
@@ -170,7 +170,7 @@ function _G.pac_Restart()
 						local str = file.Read("addons/" .. dir .. "/lua/" .. new_path, "MOD")
 						if str then
 							local func = CompileString(str, "addons/" .. dir .. "/lua/" .. new_path)
-							if type(func) == "function" then
+							if isfunction(func) then
 								local res = {pcall(func, ...)}
 
 								if res[1] then
--- a/lua/pac3/core/client/class.lua
+++ b/lua/pac3/core/client/class.lua
@@ -91,10 +91,10 @@ do
 		pac.VariableOrder[tbl.ClassName] = pac.VariableOrder[tbl.ClassName] or {}
 		insert_key(pac.VariableOrder[tbl.ClassName], key)
 
-		if type(def) == "number" then
+		if isnumber(def) then
 			tbl["Set" .. key] = tbl["Set" .. key] or function(self, var) self[key] = tonumber(var) end
 			tbl["Get" .. key] = tbl["Get" .. key] or function(self) return tonumber(self[key]) end
-		elseif type(def) == "string" then
+		elseif isstring(def) then
 			tbl["Set" .. key] = tbl["Set" .. key] or function(self, var) self[key] = tostring(var) end
 			tbl["Get" .. key] = tbl["Get" .. key] or function(self) return tostring(self[key]) end
 		else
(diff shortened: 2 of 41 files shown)


CapsAdmin:
I don't feel my opinion on this has changed much since https://github.com/CapsAdmin/pac3/pull/1022

It looks like this only affects cold code, so there is no noticable or maybe even measurable difference in performance. So what I feel is left is the ill conceived design decision that is the `is<typename>` functions.

The way I see it is that you have the type function which is overridden and written in C++, so this breaks any optimization the JIT compiler can do with the type function. Instead of writing the type function in Lua so that the JIT compiler can do its thing, you introduce a set of new `is<typename>` functions to "improve the performance by not pushing a string back to Lua"

There's so much performance potential with LuaJIT but they seem to make decisions as if JIT is not a thing which in turn takes us further away from the performance potential.

Besides performance reasons, I do see some appeal in using the is* functions, especially when you're dealing with custom gmod types.

But I'm too biased to make the correct judgment here.

CapsAdmin:
What happened to the performance issue with the type function since then? @thegrb93

FarukGamer:
@CapsAdmin
I am just giving my suggestion. You can adopt that as it is or just close it.
I can't say much about it, but as you know, pictures are worth a thousand words.

The functions were called 1 million times
image

Also, here is the code snippet I used.
https://gist.github.com/FarukGamer/6c347dabf1830ff4de98d32e94ff9ac2

CapsAdmin:
I appreciate the benchmark. Here is the same one in normal LuaJIT that illustrates my issue with this (to make it clear, it is not against you but more against gmod in general)

```
Type (JIT On): 1140593000.00 nanoseconds elapsed | 114.06 nanoseconds avg execution time.
isstring (JIT On): 1257782000.00 nanoseconds elapsed | 125.78 nanoseconds avg execution time.
Type (JIT Off): 1122947000.00 nanoseconds elapsed | 112.29 nanoseconds avg execution time.
isstring (JIT Off): 1259833000.00 nanoseconds elapsed | 125.98 nanoseconds avg execution time.
```

https://gist.github.com/CapsAdmin/a66e08ad9cd2bc69f9d225242095c405

FarukGamer:
I accidentally hit the delete button instead of editing the previous comment, sorry for that.

Anyways, I ran the benchmark again but with the code snippet you provided. If I'm Not Mistaken gmod uses a custom version of LuaJIT, in my opinion there are differences. This test was done on the x86-64 branch.

```
Type (JIT On): 22.781 secs.
isstring (JIT On): 12.833 secs.
Type (JIT Off): 22.566 secs.
isstring (JIT Off): 22.796 secs.
```

CapsAdmin:
Yes, I think your benchmark is correct in gmod, isstring is faster than type. I'm not disputing that.

Here's how the type function could be if gmod handled it correctly
```lua
Type (JIT On): 0.021150999999989 secs.
isstring (JIT On): 1.867684 secs.
Type (JIT Off): 0.021401999999995 secs.
isstring (JIT Off): 0.021090000000001 secs.
```
here's the glua code

https://gist.github.com/CapsAdmin/08294e3cab477a7c000fa42fe1ba8b38

This addon implements this optimization on our server https://github.com/PAC3-Server/notagain/blob/master/lua/notagain/optimizations/preinit/luaify.lua

you can try to run the benchmark if you want if you `connect pac3.info` but on the server, isstring is defined as `function isstring(x) type(x) == "string" end`

FarukGamer:
Tried it this time with the optimization of you and it seems to be indeed `type` is much faster, but problem is that you can not simply adopt this optimization to all servers without overwriting the functions. 

```
Type (JIT On): 0.03039519999993 secs.
isstring (JIT On): 2.5727531 secs.
Type (JIT Off): 0.030262199999925 secs.
isstring (JIT Off): 2.5833814 secs.
```

thegrb93:
> What happened to the performance issue with the type function since then? @thegrb93

Nothing's changed. x64 beta gmod still has terrible performance with any type function, and stable gmod still has marginal performance benefit using the istype functions.

FarukGamer:
I mean, the PR is not necessary. There is no noticable improvement except in synthetic benchmark.

thegrb93:
I've checked all the changes and they look fine. I think the only thing caps dislikes is using garry-isms, but there's already so much of pac3 using gmod functions that I think it's an ok thing to use.

PR #1195, opened 2022-04-19
Waive unnecessary writing from "pac_webaudio2_blankhtml.txt"

FarukGamer:
Honestly, I don't understand why this exists at all, you can realize it with SetHTML without unnecessarily writing it as a file.


CapsAdmin:
are you sure this works? I remember I had to do this to circumvent not being able to load files locally

FarukGamer:
Yes, tried it now. seems it is as you mean.

PR #1210, opened 2022-07-13
Fix an HTML injection exploit

brandonsturgeon:
Because the URL for this part is injected into an HTML blob, attackers can manually send a malicious part that can run arbitrary javascript code on any client that renders their PAC.

My fix updates the `pac.FixUrl` method to strip out exploitative HTML characters (that aren't valid in URLs anyway), and ensures that the `size` parameter is a number before concatenating it into the HTML blob.

This has been tested on our server and works as expected with no noticeable side effects.

Big thanks to @plally for helping me quickly investigate this.

### DIFF
--- a/lua/pac3/core/shared/http.lua
+++ b/lua/pac3/core/shared/http.lua
@@ -63,11 +63,7 @@ end
 
 function pac.FixUrl(url)
 	url = url:Trim()
-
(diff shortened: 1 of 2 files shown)


Fasteroid:
Can we please merge this on the `develop` branch already

PR #1227, opened 2022-09-12
quality of life: flipping the sign and switching hand sides

pingu7867:
minus or numpad minus + left clicking on a number (or vector component) will invert the sign between negative and positive

upon right clicking a string property, between paste and reset, a new option, "change sides" will change "right" to "left", or vice versa (but it doesn't work if the property is a part yet)


CapsAdmin:
good idea

pingu7867:
I can't find what I need. How would I make the left-right string thing try to assign a part for the property? Similarly to how when we manually write a string on a property that leads to a part, it tries to find the part with that name.

CapsAdmin:
I assume you figured it out?

Would be nice to be to flip vector positions and angles, but that's easier said than done (need to flip along some axis)

PR #1232, opened 2022-09-16
add voice_volume_scale

0x098:


thegrb93:
Hows this differ from voice_volume? Would it be better if voice_volume multiplied with the scale itself instead?

0x098:
> Hows this differ from voice_volume? Would it be better if voice_volume multiplied with the scale itself instead?

there could be use cases for not having it already scaled and vice versa

thegrb93:
Ok I guess. Can you fix the conflict?

0x098:
idk how to use github so i guess not

PR #1236, opened 2022-09-24
sending the entity draw shadow to the server

pingu7867:
Not sure if I'm allowed. Not sure that my code is adequate. It works for me. But I don't know how much the function can be abused if someone were inclined to do so (most I can think of is proxies or fast events might be an issue to spam messages to server). But here it is for your consideration.

I wanted to make the SendEntOverrideToServer generic in case there's other entity attributes that would be different serverside and clientside. None that I know of have come to my attention yet, but there's room to expand with other attributes.

### DIFF
--- a/lua/pac3/editor/server/wear.lua
+++ b/lua/pac3/editor/server/wear.lua
@@ -334,6 +334,8 @@ function pace.HandleReceivedData(ply, data)
 	data.owner = ply
 	data.uid = pac.Hash(ply)
 
+	pac.Message("Received pac data from ", ply, " with ", data.totalParts or 0, " parts")
+
 	if data.wear_filter and #data.wear_filter > game.MaxPlayers() then
 		pac.Message("Player ", ply, " tried to submit extraordinary wear filter size of ", #data.wear_filter, ", dropping.")
 		data.wear_filter = nil


CapsAdmin:
there is an api to do exactly this, see https://github.com/CapsAdmin/pac3/blob/master/lua/pac3/core/shared/entity_mutators/blood_color.lua for example

pingu7867:
wait that's not it. did it put my old solution in the main develop branch? I really suck at navigating github, how do I reverse it?

I wondered something else before proposing my new version: I'm not seeing any GetShadow or anything comparable that returns its boolean from the gmod Entity class side. blood color can be gotten but not draw shadow, apparently. I'm really not sure there's any indirect way to retrieve this information either.

Does that mean it's impossible unless we let pac force the drawshadow, meaning no proper mutation restorations (and I'll decide that the default shadow should be true)? Which comes back to my question of whether it's allowed.

pingu7867:
As I was syncing my fork to the newest develop, it apparently added some more commits I didn't make, putting them in this pull request. Is that what it should normally do? Well, at least it doesn't look like the files changed are anything different, you can see what I came up with for my new version using the mutator.

Issue #1235, opened 2022-09-22
bone follow part crashes on wear with TF2 rigs


pingu7867:
(possibly other non-valvebiped rigs too)

[bone_crash.txt](https://github.com/CapsAdmin/pac3/files/9628362/bone_crash.txt)

pingu7867:
For other information, if the follow part is outside of the player entity, no crash occurs.

I do have stack traces.


Executable: D:\games\SteamLibrary\steamapps\common\GarrysMod\bin\win64\gmod.exe

-Lua Stack Traces-
==================
  Client
    0. SetupBones - [C]:-1
      1. GetModelBones - addons/pac3-develop/lua/pac3/core/client/bones.lua:125
        2. GetBonePosAng - addons/pac3-develop/lua/pac3/core/client/bones.lua:303
          3. GetBoneMatrix - addons/pac3-develop/lua/pac3/core/client/base_movable.lua:84
            4. BuildWorldMatrix - addons/pac3-develop/lua/pac3/core/client/base_movable.lua:116
              5. GetWorldMatrix - addons/pac3-develop/lua/pac3/core/client/base_movable.lua:136
                6. GetWorldPosition - addons/pac3-develop/lua/pac3/core/client/base_movable.lua:147
                  7. BuildBonePositions2 - addons/pac3-develop/lua/pac3/core/client/parts/bone.lua:198
                    8. (null) - addons/pac3-develop/lua/pac3/core/client/parts/bone.lua:82
                      9. SetupBones - [C]:-1
                        10. GetModelBones - addons/pac3-develop/lua/pac3/core/client/bones.lua:125
                          11. GetBonePosAng - addons/pac3-develop/lua/pac3/core/client/bones.lua:303
                            12. GetBoneMatrix - addons/pac3-develop/lua/pac3/core/client/base_movable.lua:84
                              13. BuildWorldMatrix - addons/pac3-develop/lua/pac3/core/client/base_movable.lua:116
                                14. GetWorldMatrix - addons/pac3-develop/lua/pac3/core/client/base_movable.lua:136
                                  15. GetWorldPosition - addons/pac3-develop/lua/pac3/core/client/base_movable.lua:147
                                    16. BuildBonePositions2 - addons/pac3-develop/lua/pac3/core/client/parts/bone.lua:198

  Server
    *Not in Lua call OR Lua has panicked*

  MenuSystem
    *Not in Lua call OR Lua has panicked*

thegrb93:
[...]

PR #1237, opened 2022-09-25
Fix setupbones infinite loop

thegrb93:
Maybe fixes: #1235 
Not sure if its the correct solution though

### DIFF
--- a/lua/pac3/core/client/parts/model.lua
+++ b/lua/pac3/core/client/parts/model.lua
@@ -421,7 +421,6 @@ function PART:OnDraw()
 		self:DrawModel(ent, pos, ang)
 	self:PostEntityDraw(ent, pos, ang)
 
---  ent:SetupBones()
 	pac.ResetBones(ent)
 end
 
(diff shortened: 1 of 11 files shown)


thegrb93:
I think I need to covert other `ent:SetupBones()` to `pac.SetupBones` and it should be good

lua/pac3/core/client/parts/model.lua
```diff
@@ -444,10 +444,7 @@ local function ent_draw_model(self, ent, pos, ang)
 			self.obj_mesh:Draw()
 		cam_PopModelMatrix()
 	else
-		if ent.needs_setupbones_from_legacy_bone_parts then
-			ent:SetupBones()
-			ent.needs_setupbones_from_legacy_bone_parts = nil
-		end
+		pac.SetupBones(ent)
```
thegrb93, line 447: This should be deleted and the legacy bone part fixed

wrefgtzweve:
Can this be merged to master? this is still an issue that is being actively being exploited.

PR #1245, opened 2022-10-28
Prevent players from mutating entities if they're pac banned

wrefgtzweve:


thegrb93:
@wrefgtzweve could you PR it into develop too?

PR #1248, opened 2022-10-29
Backdoor improvements

wrefgtzweve:
Currently anyone with superadmin can run lua code on a server with pac. This isn't the biggest security concern but it'd be best to have another check in place to make sure that anyone that manages to get superadmin cannot run code without it being intentionally toggled on.
And this pr fixes a typo


Techbot121 (approved):
looks ok to me

PR #1250, opened 2022-10-29
Improve pac editor networking logic

wrefgtzweve:
- Adds a ratelimit on `pac_in_editor_posang` to prevent abuse, as this net message can send cl > sv > cl and could be abused.
- Stops players from sending pac editor net messages to the server when the editor isn't open, during some testing i noticed that half the server i was testing on (~30 players) where sending net messages while not a single one of them was using the editor. After i added the `pace.Active` check only 2 people send messages of which both where actually using the editor.

I've tested this live on a server and worked fine.


CapsAdmin:
Oh I just noticed you're targeting master. For further pull requests, can you target develop instead?

wrefgtzweve:
Alright, will do

PR #1254, opened 2022-11-06
Minor-improvements

wrefgtzweve:


thegrb93:
Can you revert the `if then return end` refactors? They make code less readable.

CapsAdmin:
I would think the opposite, "handle" any invalid input/state/errors first and then go on about doing what the function is meant to do.

Ideally you shouldn't have to do any "return end" if you fix the errors further up in stack, but I don't see any way to do this here.

thegrb93:
> I would think the opposite, "handle" any invalid input/state/errors first and then go on about doing what the function is meant to do.
> 
> Ideally you shouldn't have to do any "return end" if you fix the errors further up in stack, but I don't see any way to do this here.

Exactly, it shouldn't be the callee's responsibility. Also it makes adding additional functionality to functions a pain sometimes because you'll probably end up switching it back to a regular `if else`.

thegrb93:
Try avoiding including cleanup refactors in your code changes. We want to minimize the amount of change included with each commit.

wrefgtzweve:
Yeah, will do

lua/pac3/core/shared/movement.lua
```diff
@@ -1,52 +1,54 @@
...
-				ply.pac_movement = nil
-			else
-				if default[str] ~= nil then
-					local val = net.ReadType()
-					if type(val) == type(default[str]) then
-						ply.pac_movement = ply.pac_movement or table.Copy(default)
-						ply.pac_movement[str] = val
-					end
+		local cvar = movementConvar:GetInt()
+		if cvar ~= 1 or cvar ~= -1 or not hook.Run("PlayerNoClip", ply, true) then return end
```
thegrb93, line 33: This if statement is wrong.

thegrb93: I think the correct conversion is `if cvar ~= 1 and (cvar ~= -1 or not hook.Run("PlayerNoClip", ply, true)) then`

wrefgtzweve: Yeah i was just testing this, i'll fix it

thegrb93: I think you should use `hook.Run("PlayerNoClip", ply, true)~=false` instead of `not hook.Run("PlayerNoClip", ply, true)` as well.

wrefgtzweve: I think it should be `== false` because it'd return at true which we want

wrefgtzweve: Yeah after testing, `==` is what we need.

thegrb93: `if cvar == 0 or (cvar == -1 and hook.Run("PlayerNoClip", ply, true)==false) then`

wrefgtzweve: Yeah that's better and easier to read, I'll use that 👍

lua/pac3/editor/server/wear.lua
```diff
@@ -350,45 +350,22 @@ end
...
-
-		ply.pac_submit_spam = ply.pac_submit_spam + 1
-
-		if ply.pac_submit_spam >= pac_submit_limit:GetInt() then
-			if not ply.pac_submit_spam_msg then
-				pac.Message("Player ", ply, " is spamming pac_submit!")
-				ply.pac_submit_spam_msg = true
-			end
-
+	if pac_submit_spam:GetBool() and not game.SinglePlayer() and len > 64 then
```
thegrb93, line 357: The len check here needs to be in it's own `if then return end`

wrefgtzweve:
I considered doing that but i felt like there are cases where you want to rate limit something and not have chatprints, a simple fix would be having a ratelimit function and maybe a ratelimitwithprint function?

thegrb93:
Make it only do the alert if the message parameter isn't nil.

wrefgtzweve:
Yeah that'd work, ill do that.

lua/pac3/editor/server/wear.lua
```diff
@@ -354,7 +354,8 @@ local pac_submit_spam = CreateConVar('pac_submit_spam', '1', {FCVAR_NOTIFY, FCVA
 local pac_submit_limit = CreateConVar('pac_submit_limit', '30', {FCVAR_NOTIFY, FCVAR_ARCHIVE}, 'pac_submit spam limit')
 
 pace.PCallNetReceive(net.Receive, "pac_submit", function(len, ply)
-	if pac_submit_spam:GetBool() and not game.SinglePlayer() and len > 64 then
+	if pac_submit_spam:GetBool() and not game.SinglePlayer() then
+		if len < 64 then return end
```
thegrb93, line 358: Could the length check be moved up to the top?

wrefgtzweve: Sure, the original was in the if block so i just put it there

thegrb93 (approved):

PR #1259, opened 2022-11-11
Reduce globals

wrefgtzweve:
This pr aims to reduce the global table pollution by localizing variables that should've been local.

From:
image
image
The tool didn't display all the old globals, but there where more.

To:
image

### DIFF
--- a/lua/pac3/editor/client/panels/tree.lua
+++ b/lua/pac3/editor/client/panels/tree.lua
@@ -230,7 +230,6 @@ local function install_drag(node)
 				pace.RecordUndoHistory()
 				self.part:SetParent(child.part)
 				pace.RecordUndoHistory()
-				called = true
 			end
 		elseif self.part and self.part:IsValid() then
 			if self.part.ClassName ~= "group" then
@@ -240,14 +239,12 @@ local function install_drag(node)
 				self.part:SetParent(group)
 				pace.RecordUndoHistory()
 				pace.TrySelectPart()
-				called = true
 
 			else
 				pace.RecordUndoHistory()
 				self.part:SetParent()
 				pace.RecordUndoHistory()
 				pace.RefreshTree(true)
-				called = true
 			end
 		end
 
--- a/lua/pac3/editor/client/parts.lua
+++ b/lua/pac3/editor/client/parts.lua
@@ -239,6 +239,7 @@ function pace.GetRegisteredParts()
 end
 
 do -- menu
+	local trap
 	function pace.AddRegisteredPartsToMenu(menu, parent)
 		local partsToShow = {}
 		local clicked = false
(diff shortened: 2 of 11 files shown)


lua/pac3/editor/client/panels/tree.lua
```diff
@@ -230,7 +230,6 @@ local function install_drag(node)
 				pace.RecordUndoHistory()
 				self.part:SetParent(child.part)
 				pace.RecordUndoHistory()
-				called = true
```
thegrb93, line 233: Sure this isn't some badly made hack used by the caller function?

thegrb93: I don't see any other use of this var so maybe not.

lua/pac3/editor/client/parts.lua
```diff
@@ -330,7 +331,7 @@ do -- menu
 				pnl:SetImage(groupData.icon)
 			end
 
-			local trap = false
```
thegrb93, line 333: Why'd this local get moved?

wrefgtzweve: https://github.com/CapsAdmin/pac3/blob/0fe2506c40ff2e907adffeba230bbad0ae08b1f6/lua/pac3/editor/client/parts.lua#L273

In the original code it was being defined above it, thus creating a global

thegrb93: Oh ok, I see it now. So the code has been completely broken I think. Probably need to test whatever that trap var is used for.

thegrb93: The rest of the changes look fine.

wrefgtzweve: Now that im looking at it, i don't think this code ever runs? `pace.AddRegisteredPartsToMenu` is called by 2 functions `pace.OnPartMenu` and `pace.OnNewPartMenu` don't seem to be called anywhere? what the hell

wrefgtzweve: `pace.OnCreatePart` `pace.OnPartSelected` `pace.OnVariableChanged` `pace.OnPartMenu` `pace.OnNewPartMenu` `pace.OnHoverPart` don't seem to be used? i must be missing something

CapsAdmin: > `pace.OnCreatePart` `pace.OnPartSelected` `pace.OnVariableChanged` `pace.OnPartMenu` `pace.OnNewPartMenu` `pace.OnHoverPart` don't seem to be used? i must be missing something

they are used "magically" by pace.RunHook("PartSelected") or something like that

thegrb93: Probably string concatenation being used to call them `"On"..hookname`, which is dumb and needs to be removed someday if so.

wrefgtzweve: image
yup

wrefgtzweve: The `trap` move seems to work fine

PR #1263, opened 2022-11-20
Add server option to default `onuse` for clients

wrefgtzweve:
This pr adds an option for servers to force `onuse` to be on. Players can still disable this feature by setting `pac_onuse_only_override` to `1`.
Pac tends to lower fps significiantly on larger servers and i've spoken to some owners that would see this feature as greatly helpful.
This feature will be disabled by default and can be turned on by any willing server owner.


wrefgtzweve:
I added the convar file as i didn't see a better place to put the convar. Having this file will allow us to add other server convars and their callbacks here so they aren't hidden in files.

charitea:
Would definitely be useful, we'd love to have this option!

CapsAdmin:
Feature wise this is fine, the only thing that I don't want is adding this functionality to core. So just simply move the `pac.IsPacOnUseOnly()` convar initialization it to editor.

The idea behind core is that it should not have any restrictions and only serve as an api. The editor is what implements that api and adds restrictions on how it should be used.

Another thing with abstractions like splitting things into several files like convars is that I see `show_outfit_on_use.lua` as an isolated feature. If you remove the file you'll have safely removed the feature.

If you need server functionality and preserve the above paragraph I guess it should be moved to editor/shared, so you'd have to do what core does with its shared directory.

wrefgtzweve:
Which file do you think i should put the `pac.IsPacOnUseOnly()` exactly? i found it hard to get a good place to put it, same for the serverside convar.

CapsAdmin:
I don't think show_outfit_on_use.lua will ever be loaded because it would have to be included in server/init.lua

I would add a shared folder and setup https://github.com/CapsAdmin/pac3/blob/master/lua/autorun/pac_editor_init.lua to work the same way as https://github.com/CapsAdmin/pac3/blob/master/lua/autorun/pac_core_init.lua and move the code there with an if SERVER then check.

Then all of the relevant code is in a single file.

wrefgtzweve:
Yeah that might be better, i'll take a look at doing that. I feel like pac should pretty much autoload all of these files, we could possibly just make a editor loader? might be too far of a stretch.

PR #1265, opened 2022-11-25
Make animation parts use a stack

Jacbo1:
Before:  
https://user-images.githubusercontent.com/86734639/204056601-d44f08f8-e7f9-492a-bf53-909da7a1d4da.mp4  
After:  
https://user-images.githubusercontent.com/86734639/204056942-acf5d764-d383-4e03-a746-8baf5f741c77.mp4  
  
Animation parts now use a stack. When PART:OnShow() is called, it stops the current animation if there is one unless the current animation is the same PART, the part becomes the active animation, plays, and moves to the top of the stack. When PART:OnHide() is called, it stops its animation, removes itself from the stack, and plays the animation at the new top of the stack if it was the last one and if the stack is not empty. It's done in a way that will not interfere with anything outside of animation.lua.  
  
Edit: Forgot to mention, with Pac's current method, it's hard to make something like shown in the video because even if you try to use events to stop one animation and start another, if one starts and another stops on the same tick/frame/whatever, it breaks.

### DIFF
--- a/lua/pac3/core/client/parts/animation.lua
+++ b/lua/pac3/core/client/parts/animation.lua
@@ -268,15 +268,15 @@ function PART:OnShow()
 			if part ~= self then
 				part:OnStackStop()
 				part.pac_animation_stack_current = false
-			end
+				
+				if self.pac_animation_stack_contains then
+					-- Check this variable to save some perf
+					-- Remove self from stack to move to end and also prevent things from breaking because table.RemoveByValue() only removes the first instance
+					table.RemoveByValue(stack, self)
+				end
 
-			if self.pac_animation_stack_contains then
-				-- Check this variable to save some perf
-				-- Remove self from stack to move to end and also prevent things from breaking because table.RemoveByValue() only removes the first instance
-				table.RemoveByValue(stack, self)
+				table.insert(stack, self)
 			end
-
-			table.insert(stack, self)
 		end
 	else
 		ent.pac_animation_stack = {self}
--- a/lua/pac3/core/client/parts/animation.lua
+++ b/lua/pac3/core/client/parts/animation.lua
@@ -112,8 +112,8 @@ function PART:OnHide()
 	local ent = self:GetOwner()
 	if ent:IsValid() then
 		local stack = ent.pac_animation_stack
-		-- Remove self from animation stack
 		if stack then
+			-- Remove self from animation stack
 			if self.pac_animation_stack_contains then
 				table.RemoveByValue(stack, self)
 			end
--- a/lua/pac3/core/client/parts/animation.lua
+++ b/lua/pac3/core/client/parts/animation.lua
@@ -123,11 +123,21 @@ function PART:OnHide()
 			table.RemoveByValue(stack, self)
 		end
 
-		local part = stack[#stack]
-		if self.pac_animation_stack_current and part then
+		if self.pac_animation_stack_current then
 			-- This was the current animation so play the next in the stack
-			part:OnStackStart()
-			part.pac_animation_stack_current = true
+			local part = stack[#stack]
+			
+			-- Remove invalid parts
+			while part and not part:IsValid() do
+				stack[count] = nil
+				count = count - 1
+				part = stack[count]
+			end
+			
+			if part then
+				part:OnStackStart()
+				part.pac_animation_stack_current = true
+			end
 		end
 	end
 	
@@ -264,10 +274,21 @@ function PART:OnShow()
 			table.insert(stack, self)
 		else
 			-- Stop the current animation if it's not self
+			local count = #stack
 			local part = stack[count]
+			
+			-- Remove invalid parts
+			while part and not part:IsValid() do
+				stack[count] = nil
+				count = count - 1
+				part = stack[count]
+			end
+			
 			if part ~= self then
-				part:OnStackStop()
-				part.pac_animation_stack_current = false
+				if part then
+					part:OnStackStop()
+					part.pac_animation_stack_current = false
+				end
 				
 				if self.pac_animation_stack_contains then
 					-- Check this variable to save some perf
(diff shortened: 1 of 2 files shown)


CapsAdmin:
sounds interesting, but how does it affect previous outfits?

Jacbo1:
Should be unaffected. I only had a couple pacs that used non-custom animations and they were broken already from previous updates.

lua/pac3/core/client/parts/animation.lua
```diff
@@ -218,6 +247,41 @@ function PART:OnShow()
...
+-- Play animation and move to top of animation stack
+function PART:OnShow()
+	local ent = self:GetOwner()
+	if ent:IsValid() then
+		local stack = ent.pac_animation_stack
+		if stack then
+			local count = #stack
+			if count == 0 then
+				-- Empty stack
+				table.insert(stack, self)
```
DBotThePony, line 259: If you use #count operator to check if table is not empty, then you also can use it to insert new value into it (`stack[count + 1] = self`)

Jacbo1:
Your second commit is setting 2 of the booleans to false then checking them as if that didn't happen

Jacbo1:
The nuke commit is also probably unnecessary because the parts are already removed by the original function

Jacbo1:
Should be good now

DBotThePony:
> Does this make the table nonsequential? Why not use table.remove() 

no

> That loop also will never exit when the stack becomes empty because you don't check if it is 

if stack is empty, part is getting set to `stack[0]`, which is `nil`

Jacbo1:
Oh wait what is that even for? Checks if the part is not nil and invalid?

DBotThePony:
> Oh wait what is that even for? Checks if the part is not nil and invalid? I thought it was `==` not `~=`

read commit message of https://github.com/CapsAdmin/pac3/pull/1265/commits/ff30e74afa979900a92c1ad1ee0e949fd6753283

Jacbo1:
When would parts become invalid

DBotThePony:
> When would parts become invalid and would they not have OnHide() called?

Script errors. Nobody is protected against them. and if base game script errors don't straight up crash you to desktop, so we should be preprared to deal with our internal state once things go haywire.

DBotThePony:
@CapsAdmin @thegrb93 i see no issues, considering old code was competing for resources inside `OnThink` method

Merging this can be done via `git checkout develop && git pull origin/event-stack`, it has commits squashed.

thegrb93:
Can pac_animation_stack be a self contained /classtype? I don't like the spaghetti it's adding to animation part source.

Jacbo1:
Maybe but that would require more code and handling destroying the instances of the class and would require more changes to other files.

Jacbo1:
It's really only a table in the entity's table and a couple bools that help a bit with perf.

thegrb93:
A self contained class is more organized and isolates the spaghetti. I'll take a look in a little while

Jacbo1:
Feels like it'd just be unnecessary overhead and memory. How are you planning on doing it anyway because the stack table has to be part of the part's owner because it's interacted with by an arbitrary amount of animation parts and the bools have to be unique to each part

thegrb93:
Try moving all the push/pop code into something like this

```lua
local AnimStack = {
	__index = {
		push = function(self, part)
			
		end,
		pop = function(self, part)
			
		end
	},
	__call = function(meta)
		return setmetatable({
			stack = {}
		}, meta)
	end
}
setmetatable(AnimStack)
```

`ent.pac_animation_stack = AnimStack()`

Jacbo1:
It needs to be able to remove items at any index not just the end

Jacbo1:
Also all this would accomplish is replacing a table.insert() and table.RemoveByValue() with this code and adding even more code elsewhere

thegrb93:
Just move the shit that's in the part code into that class code.

Jacbo1:
Not sure where to put `ent.pac_animation_stack = AnimStack()` since if it goes in animation.lua it'll mess things up while you're doing stuff in the editor.

thegrb93 (changes requested):

thegrb93 (approved):

thegrb93:
Now just pull and make sure it's still working

Jacbo1:
I works now

lua/pac3/core/client/parts/animation.lua
```diff
@@ -2,6 +2,79 @@ local FrameTime = FrameTime
...
+						table.RemoveByValue(stack, part)
+					end
+
+					table.insert(stack, part)
+				end
+			end
+
+			part:OnStackStart()
+			part.pac_animation_stack_current = true
+			part.pac_animation_stack_contains = true
```
thegrb93, line 40: One last thing actually. These can be members of this stack object instead of members of the part. Mind doing that? It would also be good if there was `getTop = function(self)` so you could do `local top = self:getTop()`

thegrb93 (changes requested):

lua/pac3/core/client/parts/animation.lua
Jacbo1: How could these be part of the stack when they need to be unique for each part and the stack is only unique for the part's owner?

Jacbo1: And also if I move them into AnimStack.stack that defeats the entire point of the flags which is to avoid having to search the whole table when the part is potentially not in it

thegrb93: You could set `self.current_part = part` and replace stack_contains with `self.parts[part]`

thegrb93: The stack_contains optimization could be removed anyway though. It's not like we're expecting 1000 things on the stack

thegrb93:
Thanks. Looks much more readable now

Jacbo1:
That commit you made made it spew errors. `AnimStack` doesn't exist in the context of `AnimStack.__index.get()` because `AnimStack` needs to be declared before setting its table since it's local.

PR #1269, opened 2022-12-09
Add autocomplete to pac_wear_parts

Jacbo1:
Added autocomplete to pac_wear_parts and removed the need to wrap file names with spaces in quotes.
ezgif-1-8c9394c5da

### DIFF
--- a/lua/pac3/editor/client/spawnmenu.lua
+++ b/lua/pac3/editor/client/spawnmenu.lua
@@ -24,7 +24,7 @@ function(cmd, args)
 
 	-- Find files and directories
 	local files, dirs = file.Find("pac3/" .. args .. "*", "DATA")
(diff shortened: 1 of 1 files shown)


Yagira:
Breaks when you try to use `"`
```
[pac3-develop] addons/pac3-develop/lua/pac3/editor/client/spawnmenu.lua:26: bad argument #1 to 'ipairs' (table expected, got nil)
  1. ipairs - [C]:-1
   2. unknown - addons/pac3-develop/lua/pac3/editor/client/spawnmenu.lua:26 (x8)
```

Also breaks when you include a space after the file name.

Jacbo1:
Hold on there's still an issue with invalid file chars

Jacbo1:
Should be good now

PR #1273, opened 2023-01-06
Added bearing to event

Jacbo1:
Mainly useful for pacs like the southpark example where you have sprites that change depending on the viewing angle. The dot events are kinda scuffed for this because they don't ignore pitch so if you go too high or low compared to them (and stay looking at them ofc), it may stop drawing any sprites depending on how the pac is set up. They will also change if you turn your own camera while the pac wearer stays still.

The bearing events and proxy inputs return the yaw relative to their owner (changes depending on which version you use) to the client's camera position, making it much more suitable for these kinds of pacs. Events use degrees and proxies use radians sticking to how existing angle related events and proxies operate.
The `bearing` event and proxy input use the owner player's pos and eye angles.
The `parent_bearing` event uses the part's parent's pos and angles.
The `parent_bearing` proxy input is the same but uses the target part (similar to parent_scale_x()).
The `owner_bearing` proxy input is the same but uses the proxy's parented part (similar to owner_scale_x()).

These different versions allow you to for example put an invisible part on the head and enable Eye Angles so the bearing is relative to the head position if you wanted to put sprites on the head.


CapsAdmin:
can you also add an example using the events? maybe just duplicate the southpark one

Jacbo1:
Looks like I forgot the target part is automatically set. I had been testing with a part that I had already changed the target to clear it.

Jacbo1:
I think I'm gonna have to unmerge those event ones since the RootOwner checkbox doesn't actually seem to change in the code and the target part is automatically assigned

Jacbo1:
[southpark bearing.txt](https://github.com/CapsAdmin/pac3/files/10364157/southpark.bearing.txt)

Jacbo1:
My "smart" way turns out to not be smart. I think I'm just gonna ignore the z axis altogether

Jacbo1:
Alright that *should* be done now. No more fancy angle stuff. Just dot product between wearer's eye angles and direction from player to client eye pos ignoring pitch, roll, and z.

CapsAdmin:
I meant example as in adding it here https://github.com/CapsAdmin/pac3/blob/master/lua/pac3/editor/client/examples.lua

PR #1275, opened 2023-01-22
shotgun projectiles

pingu7867:
+ fixing the default attract mode which did nothing by default because it had a name mismatch

### DIFF
--- a/lua/pac3/core/client/parts/projectile.lua
+++ b/lua/pac3/core/client/parts/projectile.lua
@@ -68,7 +68,7 @@ BUILDER:StartStorableVars()
 		}
 	})
 	BUILDER:GetSet("Spread", 0)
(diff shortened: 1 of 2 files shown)


Yagira:
`loopback:send reliable stream overflow (x1175)`
I don't think allowing people to send 99999999999 net messages is a good idea 😃 

Also could probably just be named 'amount'

pingu7867:
I added some limiters and error messages on both client and server. It will send only one net message (that just has an extra UInt component for the multishot) and only allow projectiles when it's below limit, then the server decides how many are spawned. Tell me what you think.

I've decided to go back on giving it an editor_friendly name. It'll be named like particles with "number particles". But it's a small detail we can negotiate, we can change it depending on what the community thinks.

PR #1286, opened 2023-05-14
adding mass property to player movement

Advers:
adds a property to player movement in the editor, "mass," which modifies the mass of a player's physics object
changing mass also inversely scales physics damage, and *only* physics damage
works in multiplayer and singleplayer

4 convars:
pac_player_movement_allow_mass, which enables or disables all mass changing functionality
pac_player_movement_max_mass, which sets an upper limit to allowed mass, up to 50000 (a hardcoded limit to SetMass)
pac_player_movement_min_mass, which sets a lower limit to allowed mass, as low as 0 (also a hardcoded limit)
pac_player_movement_physics_damage_scaling, which enables or disables physics damage scaling

it is also still affected by pac_free_movement, so if any player is unable to change their movement, they also cannot change their mass and will not receive damage scaling

### DIFF
--- a/lua/pac3/core/shared/movement.lua
+++ b/lua/pac3/core/shared/movement.lua
@@ -1,7 +1,7 @@
 local movementConvar = CreateConVar("pac_free_movement", -1, CLIENT and {FCVAR_REPLICATED} or {FCVAR_ARCHIVE, FCVAR_REPLICATED}, "allow players to modify movement. -1 apply only allow when noclip is allowed, 1 allow for all gamemodes, 0 to disable")
 local allowMass = CreateConVar("pac_player_movement_allow_mass", 1, CLIENT and {FCVAR_REPLICATED} or {FCVAR_ARCHIVE, FCVAR_REPLICATED}, "enables changing player mass in player movement. 1 to enable, 0 to disable", 0, 1)
 local massUpperLimit = CreateConVar("pac_player_movement_max_mass", 50000, CLIENT and {FCVAR_REPLICATED} or {FCVAR_ARCHIVE, FCVAR_REPLICATED}, "restricts the maximum mass that players can use with player movement", 85, 50000)
-local massLowerLimit = CreateConVar("pac_player_movement_min_mass", 50000, CLIENT and {FCVAR_REPLICATED} or {FCVAR_ARCHIVE, FCVAR_REPLICATED}, "restricts the minimum mass that players can use with player movement", 0, 85)
+local massLowerLimit = CreateConVar("pac_player_movement_min_mass", 0, CLIENT and {FCVAR_REPLICATED} or {FCVAR_ARCHIVE, FCVAR_REPLICATED}, "restricts the minimum mass that players can use with player movement", 0, 85)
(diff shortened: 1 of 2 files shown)


lua/pac3/core/shared/movement.lua
```diff
@@ -145,6 +151,24 @@ pac.AddHook("Move", "custom_movement", function(ply, mv)
 
 	ply:SetJumpPower(self.JumpHeight)
 
+	if SERVER then
+		if allowMass:GetInt() == 1 then
+			ply:GetPhysicsObject():SetMass(math.Clamp(self.Mass, massLowerLimit:GetFloat(), massUpperLimit:GetFloat()))
+		end
+	end
+	
+	if (movementConvar:GetInt() == 1 and massDamageScale:GetInt() == 1 or (movementConvar:GetInt() == -1 and hook.Run("PlayerNoClip", ply, true) == true)) and massDamageScale:GetInt() == 1 then
```
pingu7867, line 160: (more readable form just for my sake of demonstration)
```
(
	movementConvar:GetInt() == 1
	and massDamageScale:GetInt() == 1
	
	or
	(
		movementConvar:GetInt() == -1
		and hook.Run("PlayerNoClip", ply, true) == true
	)
)
and massDamageScale:GetInt() == 1
```

You don't need to check massdamagescale twice. I would keep only the outside one because it governs everything anyway.
The expression is not wrong though.

pingu7867:
I don't see issues yet, concerning whether the code does what it says it does.

But I'll wait for other reviewers for whether there's other implications for server owners.  I have given it some thought with Adversarius, but I'll err on the side of caution and more peer review. Mass could've been in another part just as well, but I think it's fine in playermovement too.
Concerning the implications, the settings/convars are there though. I will add them to the expanded settings menu in due course, when I go work on that.

Advers:
remove the unnecessary massDamageScale, also changed the default minimum mass value to be 0 instead of 50000 (which it definitely wasn't supposed to be)

lua/pac3/core/shared/movement.lua
```diff
@@ -145,6 +151,24 @@ pac.AddHook("Move", "custom_movement", function(ply, mv)
...
+		scale_mass = 85/math.Clamp(self.Mass, math.max(massLowerLimit:GetFloat(), 0.01), massUpperLimit:GetFloat())
+	else
+		scale_mass = 1
+	end
+	
+	pac.AddHook("EntityTakeDamage", "PAC3MassDamageScale", function(target, dmginfo)
+		if (target:IsPlayer() and dmginfo:IsDamageType(DMG_CRUSH or DMG_VEHICLE)) then
+				dmginfo:ScaleDamage(scale_mass)
+		end
+	end)
```
thegrb93, line 170: This hook will be affecting every player rather than just the mass scaled player.

thegrb93 (changes requested):

pingu7867:
It's been a while. I've applied some more changes concerning this topic in my main dev version, and they're going to be subsumed into my combat update, So I'll be closing this pull request.

PR #1293, opened 2023-06-18
Update http.lua

WolfKann:
Adds in new url sub to fix new drop box link formats not downloading correctly, this should work immediately and has no direct effect on current version, could be pushed as a hotfix for the master-branch.


thegrb93:
Looks like this fix was already merged from something else

thegrb93:
Nvm I think github is being dumb

PR #1298, opened 2023-07-19
Rework submitpart and streamqueue

brandonsturgeon:
## The Problem
What got me started here was a relatively minor bug. I noticed that sometimes when applying a PAC, the message that confirms your part was applied:
```
[PAC3] Your part "blah" has been applied
```
wouldn't run for all of the parts I applied - it missed maybe 2/6 in my tests.

It's not a big deal, but while trying to figure out why, I found some opportunities to improve the current queue system.

## Explanation
The current `StreamQueue` system is set up to [queue a part if a part has already been submitted in the same frame](https://github.com/CapsAdmin/pac3/blob/master/lua/pac3/editor/server/wear.lua#L112-L118). The frame is calculated from [a `0.1`s timer](https://github.com/CapsAdmin/pac3/blob/master/lua/pac3/editor/server/wear.lua#L12-L18) which also runs `SubmitPart` with the queued data.

The problem here is that when `SubmitPartNotify` calls `SubmitPart`, it might receive `queue` in response if it wasn't the first part of the frame.
This means that the alert telling the client that their part was applied [wouldn't be called](https://github.com/CapsAdmin/pac3/blob/master/lua/pac3/editor/server/wear.lua#L307).

## The Changes
The simplest way to solve this would be to simply remove the:
```lua
if allowed == "queue" then return end
```

But why not just have a real queue? It could handle its own timing (no need to track an arbitrary frame count), it would work the same way every time you call it, and it keeps the logic a little tidier (queue logic in one place, "work" logic in another).

This PR turns the `StreamQueue` into the primary way of actually submitting the part.

- `SubmitPart` now adds the item to the `StreamQueue`.
- `SubmitPartNotify` calls `SubmitPart` with a callback which will perform the notifications.
- `SubmitPart` will takes a third parameter; a callback that will be called when the queue item completes. It takes `allowed, reason`, the same return values as `SubmitPartNow`.

The Diff sucks so here's a simplified diff showing, conceptually, what changed:
```diff
- function SubmitPart( data, filter )
-     if submittedThisFrame then
-         addToQueue( data, filter )
-         return "queue"
-     end
-
-   submitPartLogic()
- end
-
+ function SubmitPart( data, filter, callback )
+     addToQueue( data, filter, callback )
+ end
+
+ function SubmitPartNow( data, filter )
+    submitPartLogic()
+ end
```

The Queue timer was changed to call `SubmitPartNow`.

### Bonus Changes
I also took this opportunity to make a few styling/nitpick changes in the same area:
- Simplify some table accessors (localizing)
- Localize a few frequently called global functions (enjoy those extra microseconds, boys)
- Fix a typo or two
- Remove some dead code that complicates an already-complicated function
- Moved a few return-early checks further up the logic chain, short-circuiting earlier (see annotations on the code)

## Things to consider

This is **mostly a technical change** with a few potential downsides. I know that doesn't sound like a great deal, but the legacy PAC code could use some love. Figured I'd take a stab at helping.

If these change are too dramatic for too little a benefit, I understand. I'd be fine doing a simple 1-line approach instead.

### Bypassing the queue
Previously it looks like there were a few ways to bypass the queue. If the item wasn't really a part, or if it was empty or something? I don't fully understand it, but that won't happen here.

If there are some things that should skip the queue for performance reasons, would it be worth re-complicating the code to account for them? We run 10 queue items per second. That's pretty quick still.

Additionally, maybe we could add a convar to decide how fast the `StreamQueue` is processed?

### DIFF
--- a/lua/pac3/editor/server/wear.lua
+++ b/lua/pac3/editor/server/wear.lua
@@ -3,33 +3,35 @@ local type = type
 local istable = istable
 local IsValid = IsValid
 local tostring = tostring
+local ProtectedCall = ProtectedCall
 
 pace.StreamQueue = pace.StreamQueue or {}
 
-local function startQueueTimer()
-	timer.Create("pac_check_stream_queue", 0.1, 0, function()
-		if pace.BusyStreaming then return end
+timer.Create("pac_check_stream_queue", 0.1, 0, function()
+	if pace.BusyStreaming then return end
 
-		local item = table.remove(pace.StreamQueue)
-		if not item then return end
+	local item = table.remove(pace.StreamQueue)
+	if not item then return end
 
-		local data = item.data
-		local filter = item.filter
-		local callback = item.callback
+	local data = item.data
+	local filter = item.filter
+	local callback = item.callback
 
-		local allowed, reason = pace.SubmitPartNow(data, filter)
-		if isfunction(callback) then
+	local allowed, reason
+	local function submitPart()
+		allowed, reason = pace.SubmitPartNow(data, filter)
+	end
+
+	local success = ProtectedCall(submitPart)
+
+	if isfunction(callback) then
+		if success then
+			callback(false, "Unexpected Error")
+		else
 			callback(allowed, reason)
 		end
-	end)
-end
-
-timer.Create("pac_check_stream_queue_timer", 0.2, 0, function()
-	if not timer.Exists("pac_check_stream_queue") then
-		startQueueTimer()
 	end
 end)
-startQueueTimer()
 
 local function make_copy(tbl, input)
 	if tbl.self.UniqueID then
--- a/lua/pac3/editor/server/wear.lua
+++ b/lua/pac3/editor/server/wear.lua
@@ -1,21 +1,35 @@
+local next = next
+local type = type
+local istable = istable
+local IsValid = IsValid
+local tostring = tostring
+
 pace.StreamQueue = pace.StreamQueue or {}
 
-local frame_number = 0
-local last_frame
-local ERROR_COLOR = Color(228, 37, 37)
+local function startQueueTimer()
+	timer.Create("pac_check_stream_queue", 0.1, 0, function()
+		if pace.BusyStreaming then return end
 
-local function catchError(err)
-	pac.Message(ERROR_COLOR, 'Error: ', err)
-	pac.Message(debug.traceback())
+		local item = table.remove(pace.StreamQueue)
+		if not item then return end
+
+		local data = item.data
+		local filter = item.filter
+		local callback = item.callback
+
+		local allowed, reason = pace.SubmitPartNow(data, filter)
+		if isfunction(callback) then
+			callback(allowed, reason)
+		end
+	end)
 end
 
-timer.Create("pac_check_stream_queue", 0.1, 0, function()
-	if not pace.BusyStreaming and #pace.StreamQueue ~= 0 then
-		xpcall(pace.SubmitPart, catchError, unpack(table.remove(pace.StreamQueue)))
+timer.Create("pac_check_stream_queue_timer", 0.2, 0, function()
+	if not timer.Exists("pac_check_stream_queue") then
+		startQueueTimer()
 	end
-
-	frame_number = frame_number + 1
 end)
+startQueueTimer()
 
 local function make_copy(tbl, input)
 	if tbl.self.UniqueID then
@@ -85,7 +99,7 @@ duplicator.RegisterEntityModifier("pac_config", function(ply, ent, parts)
 
 	-- give source engine time
 	timer.Simple(0, function()
-		for uid, data in pairs(parts) do
+		for _, data in pairs(parts) do
 			if istable(data.part) then
 				make_copy(data.part, id)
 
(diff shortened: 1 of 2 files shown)


lua/pac3/editor/server/wear.lua
```diff
@@ -1,21 +1,35 @@
...
-	pac.Message(debug.traceback())
+		local data = item.data
+		local filter = item.filter
+		local callback = item.callback
+
+		local allowed, reason = pace.SubmitPartNow(data, filter)
+		if isfunction(callback) then
+			callback(allowed, reason)
+		end
+	end)
```
brandonsturgeon, line 24: This is the new primary queue timer.

It calls the given callback with `SubmitPartNow`'s return values.

Please note: This doesn't catch errors, so errors will be halting (but the timer will still be rescued).

This isn't _that_ different from how it was before, though (It just printed an error with a debug trace).
Also, because use `table.remove`, there's no chance of getting stuck on a bad item.

The only problem here is that an error in `SubmitPartNow` would prevent the callback from being called. If we need the callback to be guaranteed, we can go back to using `xpcall` (or similar). Though, I believe that JIT isn't able to optimize functions run across protected call boundaries.

If that's true, this function would likely benefit a lot from JIT optimizations. That's why I did it - based only on an unproven years-old email from one of the JIT devs 😁
So I'd be fine changing it if anyone feels strongly about it 👍

lua/pac3/editor/server/wear.lua
```diff
@@ -1,21 +1,35 @@
...
-	if not pace.BusyStreaming and #pace.StreamQueue ~= 0 then
-		xpcall(pace.SubmitPart, catchError, unpack(table.remove(pace.StreamQueue)))
+timer.Create("pac_check_stream_queue_timer", 0.2, 0, function()
+	if not timer.Exists("pac_check_stream_queue") then
+		startQueueTimer()
 	end
-
-	frame_number = frame_number + 1
 end)
+startQueueTimer()
```
brandonsturgeon, line 32: This timer exists solely as a rescue timer, to keep the main queue timer running even if it errors.

lua/pac3/editor/server/wear.lua
```diff
@@ -85,7 +99,7 @@ duplicator.RegisterEntityModifier("pac_config", function(ply, ent, parts)
 
 	-- give source engine time
 	timer.Simple(0, function()
-		for uid, data in pairs(parts) do
+		for _, data in pairs(parts) do
```
brandonsturgeon, line 102: Unused

lua/pac3/editor/server/wear.lua
```diff
@@ -108,68 +122,66 @@ duplicator.RegisterEntityModifier("pac_config", function(ply, ent, parts)
...
 end)
 
-function pace.SubmitPart(data, filter)
-	if istable(data.part) then
-		if last_frame == frame_number then
-			table.insert(pace.StreamQueue, {data, filter})
-			pace.dprint("queuing part %q from %s", data.part.self.Name, tostring(data.owner))
-			return "queue"
-		end
+function pace.SubmitPartNow(data, filter)
```
brandonsturgeon, line 125: This is the new "worker" function. It only has a few changes, which I'll annotate below.

lua/pac3/editor/server/wear.lua
```diff
@@ -108,68 +122,66 @@ duplicator.RegisterEntityModifier("pac_config", function(ply, ent, parts)
...
+	local part = data.part
+	local owner = data.owner
+
+	-- last arg "true" is pac3 only in case you need to do your checking differently from pac2
+	local allowed, reason = hook.Run("PrePACConfigApply", owner, data, true)
+	if allowed == false then return allowed, reason end
```
brandonsturgeon, line 131: I moved this `allowed == false` check up to where we run the hook. I don't understand why we ran the next block of code first, it doesn't seem necessary.

Please correct me if I'm wrong on that.

lua/pac3/editor/server/wear.lua
```diff
@@ -108,68 +122,66 @@ duplicator.RegisterEntityModifier("pac_config", function(ply, ent, parts)
...
+
+	local uid = data.uid
+	if uid ~= false and pace.IsBanned(owner) then
+		return false, "you are banned from using pac"
 	end
```
brandonsturgeon, line 136: I moved this banned check up farther too, seems like something we'd want to return early for.

lua/pac3/editor/server/wear.lua
```diff
@@ -178,26 +190,8 @@ function pace.SubmitPart(data, filter)
 
 				pace.dupe_ents[key] = nil
 			end
-		elseif data.part then
-			pace.Parts[uid][data.part] = nil
-
-			-- this doesn't work because the unique id is different for some reason
```
brandonsturgeon, line 184: Big chunk of deadcode. This function is complicated enough, there's no reason to keep it. If someone wants to come back later and change it, this comment won't help them either.

lua/pac3/editor/server/wear.lua
```diff
@@ -276,19 +272,19 @@ function pace.SubmitPart(data, filter)
...
+				if owner and owner:IsValid() then
+					owner:ChatPrint('[PAC3] ERROR: Could not broadcast your outfit: ' .. errStr)
 				end
 			else
 				net.Send(players)
 			end
 		end
 
-		if istable(data.part) then
-			last_frame = frame_number
```
brandonsturgeon, line 290: Got rid of `last_frame` because we don't need it anymore.

lua/pac3/editor/server/wear.lua
```diff
@@ -298,33 +294,58 @@ function pace.SubmitPart(data, filter)
...
+			data = data,
+			filter = filter,
+			callback = callback
+		})
+
+		return "queue"
+	end
+
+	return pace.SubmitPartNow(data, filter)
+end
```
brandonsturgeon, line 311: The new `SubmitPart` function with two key differences:
1. It takes a third `callback` param
2. It will always return `"queue"` (to maintain compatibility with the old function)

So this function will never actually call the real worker function immediately.

lua/pac3/editor/server/wear.lua
```diff
@@ -298,33 +294,58 @@ function pace.SubmitPart(data, filter)
...
-			net.WriteString(data.part.self.Name or "no name")
+			net.WriteString(reason)
+			net.WriteString(partName)
 		net.Send(data.owner)
 
-		hook.Run("PACSubmitAcknowledged", data.owner, util.tobool(allowed), reason or "", data.part.self.Name or "no name", data)
+		hook.Run("PACSubmitAcknowledged", data.owner, not not allowed, reason, partName, data)
 	end
+
+	pace.SubmitPart(data, nil, callback)
```
brandonsturgeon, line 339: `SubmitPartNotify` now just calls `SubmitPart` with a notification callback.

lua/pac3/editor/server/wear.lua
CapsAdmin: Couldn't that be solved with xpcall? (I've never seen this pattern before in glua, but I can see some benefits using a rescue timer)

lua/pac3/editor/server/wear.lua
```diff
@@ -232,26 +226,28 @@ function pace.SubmitPart(data, filter)
...
 
 	if not data.server_only then
-		if data.owner:IsValid() then
-			data.player_uid = pac.Hash(data.owner)
+		if owner:IsValid() then
+			data.player_uid = pac.Hash(owner)
 		end
 
-		local players = filter or players
+		players = filter or players
```
CapsAdmin, line 233: local players?

lua/pac3/editor/server/wear.lua
brandonsturgeon: Yes, I was trying something new.
I thought it would be more performant, but after some tests, I think simpler is better here.

I went with a `ProtectedCall` solution instead 👍

lua/pac3/editor/server/wear.lua
brandonsturgeon: It's already localized up here: https://github.com/CFC-Servers/pac3/blob/fix/rework-submitpart/lua/pac3/editor/server/wear.lua#L204

Should be fine, right?

CapsAdmin: oh right

lua/pac3/editor/server/wear.lua
```diff
@@ -1,28 +1,46 @@
...
-local function catchError(err)
-	pac.Message(ERROR_COLOR, 'Error: ', err)
-	pac.Message(debug.traceback())
-end
+	local data = item.data
+	local filter = item.filter
+	local callback = item.callback
 
-timer.Create("pac_check_stream_queue", 0.1, 0, function()
-	if not pace.BusyStreaming and #pace.StreamQueue ~= 0 then
```
brandonsturgeon, line 13: Note, I removed `pace.BusyStreaming` because it doesn't seem to exist anywhere else

PR #1305, opened 2023-08-15
"spawn as props" bone scale retention 

Heatwu:
Sets the bone scale of all the bones on spawned props when "spawn as props" is used


lua/pac3/extra/server/contraption.lua
```diff
@@ -42,6 +42,10 @@ local function spawn(val,ply)
 
 	if phys:IsValid() then
 		phys:EnableMotion(false)
+		for i=0, ent:GetBoneCount() do
+			ent:ManipulateBoneScale( i, Vector(val.scale.Y,val.scale.Z,val.scale.X) )
+		end
+		print(val.scale)
```
thegrb93, line 48: Left a print

Heatwu: fixed

lua/pac3/extra/server/contraption.lua
```diff
@@ -42,6 +42,9 @@ local function spawn(val,ply)
 
 	if phys:IsValid() then
 		phys:EnableMotion(false)
+		for i=0, ent:GetBoneCount() do
```
thegrb93, line 45: Should be `ent:GetBoneCount()-1`, also why is the vector getting the x,y,z components out of order?

Heatwu: > Should be `ent:GetBoneCount()-1`, also why is the vector getting the x,y,z components out of order?

If you do that won't that make the last bone not be set?
IDK why they are out of order but re-ordering them that way works if you look at `client/contraption` I got the size and scale directly. It may have something to do with the way pac3 scale and applied to to models

thegrb93: The last bone is index `ent:GetBoneCount()-1` because they are 0 indexed. I'll need to look at why vector components are they way.

wrefgtzweve:
Bone scaling props can be used to crash servers super easy, these values should be clamped.

PR #1310, opened 2023-10-09
Add hook for autowear

Python1320:
Will be used for disabling or changing to `autowear_terrortown.txt` so ttt can have different default outfits.

### DIFF
--- a/lua/pac3/editor/client/wear.lua
+++ b/lua/pac3/editor/client/wear.lua
@@ -258,9 +258,18 @@ do
 		elseif pace.IsActive() then
 			pac.Message("not wearing autoload outfit, editor is open")
 		else
-			pac.Message("Wearing autoload...")
-			pace.LoadParts("autoload")
-			pace.WearParts()
+			local autoload_state = "autoload"
+			autoload_state = hook.Run("PAC3Autoload", autoload_state)
+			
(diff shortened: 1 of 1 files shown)


lua/pac3/editor/client/wear.lua
```diff
@@ -258,9 +258,18 @@ do
 		elseif pace.IsActive() then
 			pac.Message("not wearing autoload outfit, editor is open")
 		else
-			pac.Message("Wearing autoload...")
-			pace.LoadParts("autoload")
-			pace.WearParts()
+			local autoload_state = "autoload"
+			autoload_state = hook.Run("PAC3Autoload", autoload_state)
```
thegrb93, line 262: This could just be `local autoload_state = hook.Run("PAC3Autoload", autoload_state)`

thegrb93 (approved):
Slightly cleaner
```lua
local autoload_file = "autoload"
local autoload_result = hook.Run("PAC3Autoload", autoload_file)

if autoload_result ~= false then
	if isstring(autoload_result) then
		autoload_file = autoload_result
	end

	pac.Message("Wearing " .. autoload_file .. "...")
	pace.LoadParts(autoload_file)
	pace.WearParts()
end
```
Fine merging as is though

PR #1315, opened 2023-10-27
update master branch to latest develop

Techbot121:
I'm mainly doing this, because there has been a change with the dynamic lights and it's really annoying to deal with on the current workshop version if you, or anyone are using lights.


.github/workflows/update_workshop.yaml
```diff
@@ -0,0 +1,30 @@
...
+
+        - name: Publish to Steam Workshop
+          uses: vurv78/gmod-upload@v0.1.3
+          env:
+            STEAM_USERNAME: ${{ secrets.STEAM_NAME }}
+            STEAM_PASSWORD: ${{ secrets.STEAM_PASSWORD }}
+          with:
+            id: ${{ github.ref == 'refs/heads/master' && '104691717' || '3038093543' }}
+            changelog: ${{ github.event.head_commit.message }}
+            config: ${{ github.ref == 'refs/heads/master' && 'addon.json' || 'addon_develop.json' }}
```
CapsAdmin, line 30: This is the only thing I can think of checking, but it seems fine with the ||

wrefgtzweve (approved):

PR #1316, opened 2023-10-29
The Combat Update. Part 1 : the combat stuff

pingu7867:
for a provisory feature list, see the readme on my repository https://github.com/pingu7867/pac3

I had to split my main branch into three for ease of review, but it's still a big task and all three parts are still related. So we will not be merging until all three are approved. Let's take our time.

### DIFF
--- a/lua/pac3/core/client/part_pool.lua
+++ b/lua/pac3/core/client/part_pool.lua
@@ -44,6 +44,10 @@ local ent_parts = _G.pac_local_parts or {}
 local all_parts = _G.pac_all_parts or {}
 local uid_parts = _G.pac_uid_parts or {}
 
+function pac.getentparts() return ent_parts or _G.pac_ent_parts or {} end
+function pac.getallparts() return all_parts or _G.pac_all_parts or {} end
+function pac.getuidparts() return uid_parts or _G.pac_uid_parts or {} end
+
 if game.SinglePlayer() or (player.GetCount() == 1 and LocalPlayer():IsSuperAdmin()) then
 	_G.pac_local_parts = ent_parts
 	_G.pac_all_parts = all_parts
@@ -496,6 +500,33 @@ end
 
 function pac.GetPartFromUniqueID(owner_id, id)
 	return uid_parts[owner_id] and uid_parts[owner_id][id] or NULL
+end
+
+function pac.FindPartByPartialUniqueID(owner_id, crumb)
+	if not crumb then return NULL end
+	if not isstring(crumb) then return NULL end
+	if #crumb <= 3 then return NULL end
+	local closest_match
+	local length_of_closest_match = 0
+	if uid_parts[owner_id] then
+		if uid_parts[owner_id][crumb] then
+			return uid_parts[owner_id][crumb]
+		end
+
+		for _, part in pairs(uid_parts[owner_id]) do
+			local start_i,end_i = string.find(part.UniqueID, crumb)
+			if start_i or end_i then
+				closest_match = part
+				if length_of_closest_match < end_i - start_i + 1 then
+					closest_match = part
+					length_of_closest_match = end_i - start_i + 1
+				end
+				
+			end
+		end
+
+	end
+	return closest_match or NULL
 end
 
 function pac.FindPartByName(owner_id, str, exclude)
--- a/lua/pac3/core/client/parts/damage_zone.lua
+++ b/lua/pac3/core/client/parts/damage_zone.lua
@@ -1,0 +1,1035 @@
+local BUILDER, PART = pac.PartTemplate("base_movable")
+
+--ultrakill parryables: club, slash, buckshot
+
+PART.ClassName = "damage_zone"
+PART.Group = "advanced"
+PART.Icon = "icon16/package.png"
+
+local renderhooks = {
+	"PostDraw2DSkyBox",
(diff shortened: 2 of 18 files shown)


thegrb93:
Can you make it a single PR. The three PR idea does not make it easier to review.

pingu7867:
> Can you make it a single PR. The three PR idea does not make it easier to review.

splitting it was not my idea. techbot told me to. I'll make a unified PR if you want. I have my opinion (I was reticent because I knew that no matter how we compartmentalize the changes, it all adds up just the same) but I just want the review process to be proper and to go smoothly. I don't care much about the politics surrounding best practices. The way I see it, we all have our own opinions, but I just want to do whichever works.

> I'm guessing most of these new features are enabled with a cvar?

If you mean whether stuff is disabled by default, it depends. Some are and some aren't. Consents are opt in, so in a multiplayer server, for users who don't consent it's almost like it was disabled for them, in a sense. But yes, there are many convars to limit the combat parts. I made sure of that.

But I decided it's not my job to disable stuff. Stuff should work by default, and I place my concern for users over server owners. That's my stance. For users, it's a bad look to have features that are shown but don't work out of the box. Server owners have the responsibility to be informed about what features are in pac3/"pac4.5" (half-joke nickname) and then decide if they want to put controls or disable stuff in their server. They can do that in 5 minutes, or way less if they're not curious, and just want to disable everything. They have the right to simply see "combat parts" and immediately run to the settings menu to disable everything.

thegrb93:
The problem with introducing new parts, especially ones that may introduce new or different gameplay mechanics, is there are many servers with strict or carefully balanced gameplay that won't want the changes. Given that the majority of servers using pac3 are darkrp, it would be a huge issue if players could now push each other or fight each other after a pac update.

pingu7867:
> The problem with introducing new parts, especially ones that may introduce new or different gameplay mechanics, is there are many servers with strict or carefully balanced gameplay that won't want the changes. Given that the majority of servers using pac3 are darkrp, it would be a huge issue if players could now push each other or fight each other after a pac update.

I've had this discussion many times, so you're not telling me anything new. I maintain my position. This is not a huge issue. Server owners will react to the update quickly enough and then their roleplay will be back to normal. They've had enough time to see my update coming. Even if that's not the case, I will bring the information to them. It's not hard to read an update log, and turn off some checkboxes after an update. If they don't WANT to do it, that's not my problem.

I might come across as confrontational at times like this, and maybe what I'm arguing against is not how you feel, though, in which case I'm sorry. Maybe you just wanted to warn me just in case, that some people have such concerns, and that I should keep them in mind. But I've had this project going for several months. I've talked to people. It's beyond impossible that I'd not know about this. And in my opinion, you'd know my position if you hanged around in the pac3 discord server.

Point being, I agree to disagree, even if your stats are true, I don't care how many servers using pac3 are control freaks about it. I cannot in good conscience call restricting pac3, pac3. gmod is a sandbox game and pac3 is a tool of creativity, so that is what I'm working for.

But anyway, gamemode-specific defaults might be a decent idea, but I'm not sure how to feel about forcing a convar's state on startup, or making convars' behaviors different on different gamemodes...

> It would help if a list of the new parts and their functions was made.

That would be the readme. My plan is, it will be shown to everyone on their first startup, with another prompt menu to read later or dismiss it forever immediately after you close the DFrame. Might be controversial. But then again, many servers have MOTDs too, so I don't see how it's any different. Why not inform people when a major update happens? At least once. Then it goes away.

And there's also the builtin part tutorials I wrote, which is used by my popup system. They're not a set of professionnally-formatted wiki pages, but I believe it does the job in explaining in further detail how they work. It's accessible by custom keyboard shortcut and by part menu action. But maybe it should be easier. We could find an appropriate bind like F1 or put it on the partmenu actions so everyone can see it easily. If the help option bothers them, they can tweak both their partmenu actions and keyboard shortcuts.

thegrb93:
1) Model and size mutators are off by default
2) It's just as easy to enable features as it is to disable them
3) We aren't Microsoft; we shouldn't release an update that adds a bunch of stuff that breaks the game for the majority of users.

thegrb93:
This feels more like a pac3 extension you should release on workshop seperately. If there's stuff regular pac3 needs to make that work, we wouldn't have any problem adding it.

pingu7867:
> 1. Model and size mutators are off by default as should most of this be

They're on in sandbox, though. But we have a compromise. I've looked at the entity mutator code. I now know what needs to be done.

> 3. We aren't Microsoft; we shouldn't release an update that adds a bunch of stuff that breaks the game for the majority of users.

I mean yeah, that's why I tried to get some opinions during the course of my development and we're doing a review.

Consents are opt-in. Is that not enough so that the majority of users doesn't get directly affected?

You don't get to speculate that something will break the game if you don't specify what it is and how it'll break the game. I want concrete talk. How can it break?

I'll assume you're talking about something new, not the social balance of roleplay servers. And the biggest way things can break which I know of is the net messages' costs and networking what happens to the entities. I did some heavy stuff about that, such as:
1-rate limiter option in milliseconds to stop spams
2-entity amount limiters to prevent ultra spikes of entity networking
3-option to completely prevent net receivers from being created for disabled parts after the restart

Rest assured, I've done all I could on that front. If it's not enough, tell me. If you have more concrete ideas, please give them.

PR #1319, opened 2023-11-04
The Combat Update : Unified pull request

pingu7867:
It was recommended to me that I also make a unified pull request instead of separate ones. Further changes will be made here.


lua/pac3/core/client/base_movable.lua
```diff
@@ -184,6 +189,58 @@ function PART:CalcAngles(ang, wpos)
...
+					nearest_ent = ent
+					nearest_dist = dist
+				end
+			end
+		end
+		return self.Angles + ( nearest_ent:GetPos() + Vector(0,0,(nearest_ent:WorldSpaceCenter() - nearest_ent:GetPos()).z * 1.5) - part_ent:GetPos()):Angle()
+	end
 
 	if self.AimPart:IsValid() and self.AimPart.GetWorldPosition then
 		return self.Angles + (self.AimPart:GetWorldPosition() - wpos):Angle()
```
CapsAdmin, line 246: Sorry for beintg a bit late and overall absent. I'd rather get to a merge than bikeshedding on how to split something up.

My take on splitting up a PR is based on features, so for example these NEAREST_* "bone" positions could be one feature (don't bother to do this now though). It's a single feature it's independent on the broader combat specific code.

CapsAdmin:
I'll try to describe all the features in comments, then we can make a list when we merge and hopefully workshop will also list those features as changes.

thegrb93:
I think breaking this up would be too much work. It will take a while to review though.

CapsAdmin:
I tried to break up the small changes into commits here https://github.com/CapsAdmin/pac3/tree/capsadmin/cedric-small-changes if that helps.

lua/pac3/core/client/parts/interpolated_multibone.lua
```diff
@@ -0,0 +1,230 @@
+local BUILDER, PART = pac.PartTemplate("base_drawable")
+
+PART.ClassName = "interpolated_multibone"
```
CapsAdmin, line 3: Can we name this something like "part interpolation"? Initially I thought it had something to do with bones

lua/pac3/core/client/parts/interpolated_multibone.lua
```diff
@@ -0,0 +1,230 @@
...
+
+BUILDER:StartStorableVars()
+	:SetPropertyGroup("test")
+		:GetSet("Preview", false)
+	:SetPropertyGroup("Interpolation")
+		:GetSet("LerpValue",0)
+		:GetSet("Power",1)
+		:GetSet("InterpolatePosition", true)
+		:GetSet("InterpolateAngles", true)
+	:SetPropertyGroup("Nodes")
```
CapsAdmin, line 19: Likewise, these are parts, not nodes

CapsAdmin: Would be nice to have a single part list property, but not sure at the top of my head how that should work.

lua/pac3/core/client/base_movable.lua
```diff
@@ -184,6 +189,58 @@ function PART:CalcAngles(ang, wpos)
...
+	if pac.StringFind(self.AimPartName, "NEAREST_LIFE", true, true) then
+		local nearest_ent = self:GetRootPart():GetOwner()
+		local nearest_dist = math.huge
+		local owner_ent = self:GetRootPart():GetOwner()
+		local part_ent = self:GetOwner()
+		for _,ent in pairs(ents.GetAll()) do
+			if (ent:IsNPC() or ent:IsPlayer()) and ent ~= owner_ent then
+				local dist = (owner_ent:GetPos() - ent:GetPos()):LengthSqr()
+				if dist < nearest_dist then
+					nearest_ent = ent
```
CapsAdmin, line 237: can you create a local helper function that gets all of the nearby entities to avoid the repetition?

lua/pac3/core/client/parts/command.lua
```diff
@@ -25,7 +25,10 @@ end
 
 function PART:OnShow(from_rendering)
 	if not from_rendering and self:GetExecuteOnShow() then
-		self:Execute()
+		timer.Simple(0, function()
+			if self.Hide or self:IsHidden() then return end
```
CapsAdmin, line 29: am i correct thinking that this does not execute the command when hidden from rendering, but does execute when hidden with event?

lua/pac3/core/client/parts/light.lua
```diff
@@ -68,6 +83,10 @@ function PART:OnDraw()
 	self:GetLight().pos = pos
 	self:GetLight().dir = ang:Forward()
 end
+function PART:SetStyle(val)
+	self.Style = val
+	self:GetLight().Style = self.Style
```
CapsAdmin, line 88: I'm not sure how useful this is. You can mimic a light style by just adding a proxy?

lua/pac3/core/client/parts/lock.lua
```diff
@@ -0,0 +1,523 @@
+include("pac3/extra/shared/net_combat.lua")
```
CapsAdmin, line 1: Why do you include this here? same goes for other places where you add include. It will rerun all the net_combat code for every inclusion, and it might cause issues with local state if net_combat.lua or any other file has it.

It should be enough to include it once in init before parts are loaded.

lua/pac3/editor/client/popups_part_tutorials.lua
```diff
@@ -0,0 +1,1180 @@
...
+				pace.PartTemplates[i].TutorialInfo = v
+			end
+		end
+	end
+end
+
+local motd_cvar = CreateConVar("pac_show_message_on_startup", "2", {FCVAR_ARCHIVE}, "Whether to show the update MOTD when you load in the game")
+
+
+function pac.OpenMOTD(from_initial_startup)
```
wrefgtzweve, line 1146: I don't think this is needed, you're forcing everyone a extra MOTD in their face and then requiring a second popup to confirm if they've read it. This should probably be removed

lua/pac3/core/client/parts/interpolated_multibone.lua
pingu7867: I can't be the only one who started using these. Changing the classname is a big decision to make.

Maybe use a FriendlyName field to rename it? But the idea with multibone is that it can ultimately move between bones. Bones being the origin of all base_movables, it was a serious limitation that we couldn't smoothly shift a part between them. The choice of name doesn't really matter though, but undoubtedly the goal is to get the point across.

here's more ideas. tell me if one strikes your fancy.
path
interpolator
position_morpher

lua/pac3/core/client/parts/interpolated_multibone.lua
pingu7867: We can rename these with editor_friendly fields no problem. But yes, in my opinion those are nodes. Nodes are points, as in a graph. That's what a base_movable part can do if we take its position. But that's just linguistic talk. It doesn't change how the code works.

As for making it dynamic, I know it would've been nice. But I decided not to mess around very much in the deeper parts of the pac code like dynamic properties.

lua/pac3/core/client/base_movable.lua
pingu7867: for readability or for performance?

lua/pac3/core/client/parts/command.lua
pingu7867: I don't think so. But idk, I haven't thought that far. There was an issue where it wrongly auto-triggered with timerx events or something. In the gmod wiki, I read a few times that a 0 timer can be used for various things. Therefore I tried that and I think it worked.

lua/pac3/core/client/parts/light.lua
pingu7867: Those are a source engine feature and they were in the legacy lights.

No, the styles aren't easily replicated with proxies. They're arbitrary sequences of levels, most of them are not a nice continuous function. Also it's easier to setup something quick that way.

lua/pac3/core/client/parts/lock.lua
pingu7867: I don't actually remember that one. Why indeed?

lua/pac3/editor/client/popups_part_tutorials.lua
pingu7867: I knew it was gonna be controversial but I simply saw fit that, given how big the additions were, there should be a more direct way to inform server owners and users, because the principle of proportionality. I have a better solution though. I'll just use the gmod's AddLegacy notifications in the corner. You're right. Even if it's just once, it's a bother.

CapsAdmin:
Another small issue is the MOTD url. It should point to a markdown file in this repo, so just add it to your branch and fix the url.

pingu7867:
> Another small issue is the MOTD url. It should point to a markdown file in this repo, so just add it to your branch and fix the url.

I didn't want to overwrite your readme here, and my branch is still in my repo. What's the problem with getting the update log from my repository? Either I'm being difficult or you weren't clear enough; while I get the idea, I'm not sure what specific things you want me to do, and where.

So I should make a new md file to put my update's readme, and it'll eventually be brought in to your repository and I'll fix the url after?

CapsAdmin:
> So I should make a new md file to put my update's readme, and it'll eventually be brought in to your repository and I'll fix the url after?

Yes just make a new markdown file called changelog.md or something and point the url to that.

For example with a markdown file: https://github.com/CapsAdmin/pac3/blob/develop/CONTRIBUTING.md

Not sure how to go about selecting develop or master though. But this points to a larger problem, there are 2 versions of pac now, master and develop. Do you make 2 changelogs for each? If you point it at the develop version then people on master will see changes for the develop version.

Typically changelogs are made for the final release, in this case master which is then automatically deployed to workshop.

pingu7867:
> But this points to a larger problem, there are 2 versions of pac now, master and develop. Do you make 2 changelogs for each?

We do not. Master governs. Also, if we make a changelog, we will not be updating it unless there's changes of significance and magnitude. It's not about final release or not, it's about what happens with the kinds of people who are gonna read it. The people who know are usually up to date, they will go get the commit history. The average people will be content with the big changes. Less time, more important information.

The only problem happens to be the other thing with average people, a LOT of people (who got told or guided to installing a github install) don't realize their local version is not updated automatically. It's not a problem yet, but if they eventually decide to update their version (to one where the changelog would be present and linked) and not touch it for months or years, they'll have outdated information if the hosted changelog gets updated. Workshop is the only way to ensure the actual local code run corresponds. But it has its problems (the regular ban incidents).
 
My view is that in our case the information absolutely must be local to reflect what they have on their machine; because we don't have the means to prevent all people from forgetting that, if they use a local install, they must keep up with updates manually. But well, it's just some thoughts. I don't have to design the perfect information system.

Now, because it's a major update and my passion project, if you allow me a personal choice, I would still put my readme in. I did write it in a fairly informal tone but it gets most of the features fairly concisely. We can reformat it if need be. I'll show you a prototype in due course. But relating to the previous points, I'm putting it as a HTML string in the code rather than sourcing it from the internet. I don't know of a way to locally load the current branch's md file when ingame, so for now we can go with HTML strings, if that's not too crazy. I'll probably make a new lua file to make space for my readme and any further official changelog.

CapsAdmin:
> > But this points to a larger problem, there are 2 versions of pac now, master and develop. Do you make 2 changelogs for each?
> 
> We do not. Master governs. Also, if we make a changelog, we will not be updating it unless there's changes of significance and magnitude. It's not about final release or not, it's about what happens with the kinds of people who are gonna read it. The people who know are usually up to date, they will go get the commit history. The average people will be content with the big changes. Less time, more important information.
> 
> The only problem happens to be the other thing with average people, a LOT of people (who got told or guided to installing a github install) don't realize their local version is not updated automatically. It's not a problem yet, but if they eventually decide to update their version (to one where the changelog would be present and linked) and not touch it for months or years, they'll have outdated information if the hosted changelog gets updated. Workshop is the only way to ensure the actual local code run corresponds. But it has its problems (the regular ban incidents).
> 
> My view is that in our case the information absolutely must be local to reflect what they have on their machine; because we don't have the means to prevent all people from forgetting that, if they use a local install, they must keep up with updates manually. But well, it's just some thoughts. I don't have to design the perfect information system.
> 
> Now, because it's a major update and my passion project, if you allow me a personal choice, I would still put my readme in. I did write it in a fairly informal tone but it gets most of the features fairly concisely. We can reformat it if need be. I'll show you a prototype in due course. But relating to the previous points, I'm putting it as a HTML string in the code rather than sourcing it from the internet. I don't know of a way to locally load the current branch's md file when ingame, so for now we can go with HTML strings, if that's not too crazy. I'll probably make a new lua file to make space for my readme and any further official changelog.

Having the changelog checked in sounds good. Inline html is fine.

lua/pac3/core/client/parts/interpolated_multibone.lua
```diff
@@ -0,0 +1,230 @@
...
+		self.pos = self:GetWorldPosition()
+	end
+	if not self.InterpolateAngles then
+		self.ang = self:GetWorldAngles()
+	end
+	self.Owner:SetPos(self.pos)
+	self.Owner:SetAngles(self.ang)
+end
+
+function GetClosestAngleMidpoint(a1, a2, proportion)
```
thegrb93, line 142: Angle interpolation should be a rotation axis between the two angle endpoints of length (rotation radians * proportion). We shouldn't be interpolating yaw pitch roll values. I don't remember if glua has a slerp function that can do this but sad if not. If you need code for this lemme know.

thegrb93 (changes requested):

lua/pac3/core/client/parts/interpolated_multibone.lua
pingu7867: I think I just got ahead of myself out of wanting to implement it myself, it didn't end up working perfectly because of roll and some oddities, but it was something. I will use LerpVector and LerpAngle.

CapsAdmin:
My overall position on the "on by default" stuff in sandbox or sandbox derived gamemodes is that if you ask for sufficient permissions it should be OK. In other gamemodes, it should probably be off by default.

Some examples: 

* if you are allowed to noclip, then teleporting is probably fine. Worst case you could make it so you can't teleport through walls or can't teleport very far or something if there are additional restrictions to noclip.

* If you're allowed to damage another player, then they should be able to receive damage from pac too as long as it comes from you.

* If you're allowed to use the toolgun to color another player, then you should probably also be allowed to add outfits to the players.

There's no straight forward way of checking these things. It's a combination of making sure damage comes from the right player, the right weapon, location, etc, while maybe also testing if some hooks can run.

An ideal example for shooting projectiles at another player: It should appear to all other scripts as if you're spawning a prop and throwing it at a player. When this is done correctly then any anti prop push scripts, prop protection, damage protection, etc should do their job.

I think some of pac works like this, but if not it probably should. However this is my position on this point.

lua/pac3/core/client/base_movable.lua
pingu7867: > gets all of the nearby entities
That's... just ents.findinsphere or findinbox. it already exists. although I used ents.getall in this case. But perhaps it should pick the sphere one and for it to be limited to a certain radius. I'll pick 5000 or something.

The only repetition is the processing that comes after, to get the nearest entity. I'll compact the loops into that helper function you wanted.

lua/pac3/core/client/parts/light.lua
pingu7867: > if they were in legacy and not in the new, that means I didn't observe any purpose for them and let them out to reduce the amount of properties. They are also not smoothly played back as far as I can remember, feeling very hl1-like.

Well, I see a purpose. But now we care about the amount of properties? You didn't seem to object when there were other parts I've added more stuff to. I added data sources to the text part, a lot of them are single-liner functions but we want them for ease of setting up.

And I'm one of the proxy enthusiasts, you'd think I'd want more people using proxies. But these are not extra steps we need.

I think people like having easier options that are readily available. And I'm not a data guy but roughly judging from the discord's help channel, most people don't use proxies or know much about them. So I see ample reason to make things easier for users. Anyone with a light can easily get something nice in two clicks.

> Perhaps one idea is a proxy function that allows you to loop a sequence of numbers, a bit like the interpolator.

New proxy function ideas? Now that's right up my alley.

lua/pac3/core/shared/entity_mutators/draw_shadow.lua
```diff
@@ -0,0 +1,23 @@
+local MUTATOR = {} 
+  
+MUTATOR.ClassName = "draw_shadow"
```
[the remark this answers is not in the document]
pingu7867: That is a leftover file from a bad pull request of more than a year ago when I was lost and clueless about how to use github. The pull request branch ended up lost (force-pushed) and its history was erased but the file lingered in my local files.

I will go re-implement it.

materials/icon64/new pac icon.png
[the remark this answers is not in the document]
pingu7867: Good question. What does it look like? A rebranding attempt, or a joke easter egg for people to have a chuckle at?

I thought it'd be obvious with how goofy my icon looks. Garry's Mod is a game well-known for its general identity of not taking itself too seriously. So I had a little fun making a funny icon for people who choose to be in on the joke. It's off by default. I'm not changing "the" logo.

But there's a portion of truth in advancing the pac version. You may not agree because of technicalities, if you're of the opinion that the pac number should only be changed when the deep and fundamental code is majorly rewritten.

But I'm thinking of major features and I notice that we've come a long way, even pre-combat era.
-Web resources. MDL zips. You mean I can use pretty much any image or model that I like in my pac? Nice! Those are already big enough changes, which is why at that time, it was semi-jokingly called PAC4 already. The idea stuck with many people because we're not the type of community where the jokes are made by coming up with ridiculous and blatant falsehoods. We're more the caricaturing exaggeration types who always start with a grain of truth.

But now the argument is stronger.
-Combat parts. Controversial features for certain gamemodes, but no matter. Doesnèt change the fact that combat parts really do flesh out the new dimension of pac (combat) that saw its beginnings with projectiles. More ways to have fun and do cool stuff.
-Huge amounts of editor features and customization. Custom shortcuts really can change how you use the editor. Of course, we're only seeing the beginning, I don't know how much it will be used. People will see what they can do with it. But no matter how you look at it, it's a major set of features, given how the editor is the beating heart of pac, that people use every day.

Also, I take into account the drip-feed of small features and fixes adding up over time. When combined, they are worth just as much; they just take more time to get there one bit at a time. So I don't care if there's no clear-cut way of deciding how to call pac's current state. They do contribute to pac's version number, if there is such a thing.

CapsAdmin:
I feel like you've responded to a lot of the feedback, perhaps not everything. But if no one has time to review I want to merge this to develop to get more feedback.

From now on just commit directly to develop to fix any issues related to this PR.

PR #1339, opened 2024-02-24
Fixed Error regarding missing/removed Color-Metatable

Benn20002:
Since Garry's Mod seems to have removed the ability to retrieve the color metatable, use the color metafunctions directly instead.


thegrb93:
] lua_run print(FindMetaTable("Color"))
> print(FindMetaTable("Color"))...
table: 0xc0eb7a72
] lua_run_cl print(FindMetaTable("Color"))
table: 0xcdf97aa2

Something seems to be wrong with your gmod

PR #1347, opened 2024-04-12
Add size limit to E2 functions

Denneisk:
PAC3's E2 functions do not have limits on the amount of data put into them, meaning you can very easily overflow an entire server's clients. This PR adds a limit to this, and also includes small changes to fit modern Wiremod practices.

- Add size limit in bytes for PAC E2 functions, default 8192 bytes, resets every tick
- Replaces `E2Lib.raiseException` with `self:throw()` (functionally identical)
- Removes wrapping vector and angle types (unnecessary for newer version of Wiremod)


thegrb93:
Shouldn't the RatelimitPlayer function suffice?

thegrb93:
It needs to be a global rate limit because spamming 65kB will also net buffer overflow eventually.

Denneisk:
This is the only non-streamed transfer from server to client. Is it worth being concerned if the player (most likely intentionally) tries to overload his connection to the server?

Denneisk:
As for the second part, that would require a bunch of players collaborating at once to overload everyone at the same time. Not impossible but seems unlikely.

thegrb93:
I guess 65kB per tick is ok

thegrb93 (approved):

thegrb93:
Can you rebase on develop and make a PR to the develop branch?

Denneisk:
> I guess 65kB per tick is ok. They'll only get 300 ticks before their tick rate limit is hit.

I think you misread, the limit defaults to 8192. I'm actually going to lower it to 2048 since, although I don't know *what* this is used for, it's probably not used for long strings most of the time, and the other 3 datatypes are peanuts compared to 2048.

PR #1348, opened 2024-04-13
Add size limit to E2 functions

Denneisk:
See #1347

- Changed description on `pac_e2_ratelimit_buffer` and `pac_e2_ratelimit_refill` to be more descriptive
- Fixed ratelimit used `GetInt` instead of `GetFloat` causing instant limit hits with values below 1

----
PAC3's E2 functions do not have limits on the amount of data put into them, meaning you can very easily overflow an entire server's clients. This PR adds a limit to this, and also includes small changes to fit modern Wiremod practices.

- Add size limit in bytes for PAC E2 functions, default 2048 bytes, resets every tick
- Replaces `E2Lib.raiseException` with `self:throw()` (functionally identical)
- Removes wrapping vector and angle types (unnecessary for newer version of Wiremod)

### DIFF
--- a/lua/entities/gmod_wire_expression2/core/custom/pac.lua
+++ b/lua/entities/gmod_wire_expression2/core/custom/pac.lua
@@ -1,14 +1,11 @@
 E2Lib.RegisterExtension("pac", true)
 
-util.AddNetworkString("pac_e2_setkeyvalue_str")
-util.AddNetworkString("pac_e2_setkeyvalue_num")
-util.AddNetworkString("pac_e2_setkeyvalue_vec")
-util.AddNetworkString("pac_e2_setkeyvalue_ang")
+util.AddNetworkString("pac_e2_setkeyvalue")
 
 local enabledConvar = CreateConVar("pac_e2_ratelimit_enable", "1", {FCVAR_ARCHIVE}, "If the e2 ratelimit should be enabled.", 0, 1)
 local rate = CreateConVar("pac_e2_ratelimit_refill", "0.025", {FCVAR_ARCHIVE}, "The speed at which the ratelimit buffer refills.", 0, 1000)
 local buffer = CreateConVar("pac_e2_ratelimit_buffer", "300", {FCVAR_ARCHIVE}, "How large the ratelimit buffer should be.", 0, 1000)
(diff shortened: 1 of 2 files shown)


thegrb93 (approved):

Denneisk:
> Even 8k bytes per tick can stack up, gmod is limited to 64kb/s anything over that will cause choke, once you enter a buffer size of ~128kb players will start timing out/getting kicked.
> 
> The limit should be per second to accommodate for this so it'll work on any tickrate.

I updated the description to mention it's now 2048 bytes, but I'll take your suggestion for making it based on seconds.
 
> The net messages should also be set to unreliable so players don't time out.

Maybe? I guess it's no big deal if these are dropped but any point for desync seems undesirable.

wrefgtzweve:
Unreliable doesn't make the messages randomly drop, they're still equally reliable as regular net messages.
The only difference is that once the buffer is full, instead of kicking players it insteads drops the excess messages, see https://wiki.facepunch.com/gmod/Networking_Usage#unreliablechannel

thegrb93:
unreliable does induce a chance that packet loss will also drop the message, but this is rarer with modern internet. The reason for dropping is usually because of hitting the buffer limit but there are other rarer causes like ip header corruption and such

Denneisk:
> Unreliable doesn't make the messages randomly drop, they're still equally reliable as regular net messages.

My concern is that these relatively small (a typical message is going to be: 3.5B entities+type, 64B ID, ~5-20B key, 8-12B non-string value) PAC3 messages would get wiped because of the totality of unreliable dropping, but whatever, it's probably nothing to worry about in most scenarios.

> reliable uses handshakes to make sure all messages are received,

I wasn't aware if there was any extra work done on between the two, my understanding of `net` is that reliability only controls whether a message is allowed to be dropped or not.

thegrb93:
> I wasn't aware if there was any extra work done on between the two, my understanding of `net` is that reliability only controls whether a message is allowed to be dropped or not.

I don't think so because unreliable messages are noticeably much lower latency

Denneisk:
Including fixes for some bugs I encountered while testing.

thegrb93 (approved):

PR #1350, opened 2024-04-18
Fix "fire_bullets" event from erroring.

unknao:
Made the fire_bullets event functional again.

### DIFF
--- a/lua/pac3/core/client/parts/event.lua
+++ b/lua/pac3/core/client/parts/event.lua
@@ -1462,8 +1462,8 @@ PART.OldEvents = {
 		operator_type = "string", preferred_operator = "find simple",
 		tutorial_explanation = "fire_bullets supposedly checks what types of bullets you're firing",
 		arguments = {{find_ammo = "string"}, {time = "number"}},
-		callback = function(self, ent, find, time)
-			time = time or 0.1
+		callback = function(self, ent, find_ammo, time)
+			local time = time or 0.1
 
 			ent = try_viewmodel(ent)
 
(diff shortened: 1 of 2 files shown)


lua/pac3/core/client/parts/event.lua
```diff
@@ -1462,8 +1462,8 @@ PART.OldEvents = {
 		operator_type = "string", preferred_operator = "find simple",
 		tutorial_explanation = "fire_bullets supposedly checks what types of bullets you're firing",
 		arguments = {{find_ammo = "string"}, {time = "number"}},
-		callback = function(self, ent, find, time)
-			time = time or 0.1
+		callback = function(self, ent, find_ammo, time)
+			local time = time or 0.1
```
thegrb93, line 1466: This local isn't needed

thegrb93 (approved):

lua/pac3/core/client/parts/event.lua
```diff
@@ -3382,10 +3382,6 @@ pac.AddHook("EntityFireBullets", "firebullets", function(ent, data)
 	ent.pac_fire_bullets = {name = data.AmmoType, time = pac.RealTime, reset = true}
 
 	pac.CallRecursiveOnAllParts("OnFireBullets")
-
-	if ent.pac_hide_bullets then
-		return false
-	end
```
thegrb93, line 3388: Idk if removing this is right. Someone else verify

pingu7867: At first glance, if we search "pac_hide_bullets", it seemed as though the field isn't created anywhere (so, accessing it would always be nil if that was the case)
But, if we run the search without the pac_ prefix, it leads to EntityField functions, the builder thing that set an entity field with that prefix.
(for OP's information) This one wasn't obvious, that prefix business is just a thing that we do a few times throughout the code.

We still want that return because we still want the option to hide the bullets.

But there's some side issues in singleplayer. I tried to test, it appears the EntityFireBullets hook is ineffective. But it's fine with local server and p2p setups for some reason... oh well!

thegrb93: @unknao can you revert this change?

thegrb93 (approved):

pingu7867:
That's pretty much my final commit on this PR unless there's something I should amend. Awaiting approval, and I'll merge after.

PR #1363, opened 2024-06-09
Use player.Iterator() instead of player.GetAll()

Astralcircle:
It will help to reduce the senseless number of player.GetAll() calls and reduce CPU load


TW1STaL1CKY:
I would rather this be merged into the develop branch. You can try it, but I predict there's gonna be conflicts since develop is way ahead of master right now.
Otherwise I can close this PR and apply change along with other similar replacements myself (I'll reference this PR in the commit message).

PR #1365, opened 2024-06-17
Optimisation pass

TW1STaL1CKY:
I figured I'd continue on after #1364 and look at replacing all player.GetAll calls where applicable, which led into replacing all ents.GetAll calls with ents.Iterator. That brings us to now where I figured I'd keep going with every optimisation method I could think of within reason.

Instead of pushing this straight to develop, I've made this PR to give everyone interested a chance to check I haven't missed anything obvious. I've tried usual PAC usage on my own game with these changes and it seems fine.
If there's no objections, I'll merge this within a few days.

## Commit description:
Had a scanthrough and made the following changes in the name of optimisation (even micro-optimisations) where possible:

- player.Get*  ->  player.Iterator
- ents.GetAll  ->  ents.Iterator
- ipairs  ->  for i=1,#tbl (only main uses of ipairs has been replaced, not every use)
- Vector:Distance  ->  Vector:DistToSqr
- Vector:Length  ->  Vector:LengthSqr
- net.Read/WriteEntity  ->  net.Read/WritePlayer where it is definitely intended to be a player
- Saved some GetConVar calls to local vars
- Localised/reused some Vector and Color objects
- Localised some functions where they are frequently used in a few files
- Misc minor styling cleanup (mostly spacing)
- BONUS: Fixed pac.ColorToNames being incorrect when used by some parts (some now use the new pac.VectorColorToNames func)

### DIFF
--- a/lua/autorun/netstream.lua
+++ b/lua/autorun/netstream.lua
@@ -87,7 +87,7 @@ local ReadStreamQueue = {
 				ErrorNoHalt("Receiving too many ReadStream requests!")
 				return
 			end
-
+			
 			for _, v in ipairs(queue) do
 				if v.identifier == stream.identifier then
 					ErrorNoHalt("Tried to start a new ReadStream for an already existing stream!")
@@ -296,6 +296,7 @@ local ReadingDataItem = {
 }
 setmetatable(ReadingDataItem, ReadingDataItem)
 
+
 function net.WriteStream(data, callback, dontcompress)
 	if not isstring(data) then
 		error("bad argument #1 to 'WriteStream' (string expected, got " .. type(data) .. ")", 2)
@@ -309,18 +310,20 @@ function net.WriteStream(data, callback, dontcompress)
 		data = util.Compress(data) or ""
 	end
 
-	if not data[1] then
+	if #data == 0 then
 		net.WriteUInt(0, 32)
 		return
-	elseif #data > net.Stream.MaxSize then
-		ErrorNoHalt("net.WriteStream request is too large! ", #data / 1048576, "MiB")
+	end
+
+	if #data > net.Stream.MaxSize then
+		ErrorNoHalt("net.WriteStream request is too large! ", #data/1048576, "MiB")
 		net.WriteUInt(0, 32)
 		return
 	end
 
 	local stream = net.Stream.WriteStreams:Add(WritingDataItem(data, callback, compressed))
 	if not stream then return end
-
+	
 	--print("WriteStream", #stream.chunks, stream.identifier, compressed)
 	net.WriteUInt(#stream.chunks, 32)
 	net.WriteUInt(stream.identifier, 32)
@@ -341,11 +344,10 @@ function net.ReadStream(ply, callback)
 			error("bad argument #1 to 'ReadStream' (Tried to use a NULL entity!)", 2)
 		end
 	end
-
 	if not isfunction(callback) then
 		error("bad argument #2 to 'ReadStream' (function expected, got " .. type(callback) .. ")", 2)
 	end
-
+	
 	local numchunks = net.ReadUInt(32)
 	if numchunks == nil then
 		return
--- a/lua/autorun/netstream.lua
+++ b/lua/autorun/netstream.lua
@@ -87,7 +87,7 @@ local ReadStreamQueue = {
 				ErrorNoHalt("Receiving too many ReadStream requests!")
 				return
 			end
-			
+
 			for _, v in ipairs(queue) do
 				if v.identifier == stream.identifier then
 					ErrorNoHalt("Tried to start a new ReadStream for an already existing stream!")
@@ -296,7 +296,6 @@ local ReadingDataItem = {
 }
 setmetatable(ReadingDataItem, ReadingDataItem)
 
-
 function net.WriteStream(data, callback, dontcompress)
 	if not isstring(data) then
 		error("bad argument #1 to 'WriteStream' (string expected, got " .. type(data) .. ")", 2)
@@ -310,20 +309,18 @@ function net.WriteStream(data, callback, dontcompress)
 		data = util.Compress(data) or ""
 	end
 
-	if #data == 0 then
+	if not data[1] then
 		net.WriteUInt(0, 32)
 		return
-	end
-
-	if #data > net.Stream.MaxSize then
-		ErrorNoHalt("net.WriteStream request is too large! ", #data/1048576, "MiB")
+	elseif #data > net.Stream.MaxSize then
+		ErrorNoHalt("net.WriteStream request is too large! ", #data / 1048576, "MiB")
 		net.WriteUInt(0, 32)
 		return
 	end
 
 	local stream = net.Stream.WriteStreams:Add(WritingDataItem(data, callback, compressed))
 	if not stream then return end
-	
+
 	--print("WriteStream", #stream.chunks, stream.identifier, compressed)
 	net.WriteUInt(#stream.chunks, 32)
 	net.WriteUInt(stream.identifier, 32)
@@ -344,10 +341,11 @@ function net.ReadStream(ply, callback)
 			error("bad argument #1 to 'ReadStream' (Tried to use a NULL entity!)", 2)
 		end
 	end
+
 	if not isfunction(callback) then
 		error("bad argument #2 to 'ReadStream' (function expected, got " .. type(callback) .. ")", 2)
 	end
-	
+
 	local numchunks = net.ReadUInt(32)
 	if numchunks == nil then
 		return
--- a/lua/pac3/core/client/base_part.lua
+++ b/lua/pac3/core/client/base_part.lua
@@ -13,6 +13,9 @@ local Color = Color
 local NULL = NULL
 local table_insert = table.insert
 
+local pac_editor_scale = GetConVar("pac_editor_scale")
+local pac_popups_preferred_location = GetConVar("pac_popups_preferred_location")
+
 local BUILDER, PART = pac.PartTemplate()
 
 PART.ClassName = "base"
@@ -53,6 +56,18 @@ function PART:IsValid()
 	return self.is_valid
 end
 
+local pac_enable_convars = {}
+local function get_enable_convar(classname)
+	local convar = pac_enable_convars[classname]
+
+	if not convar then
+		convar = GetConVar("pac_enable_" .. string.Replace(classname, " ", "_"):lower())
+		pac_enable_convars[classname] = convar
+	end
+
+	return convar
+end
+
 function PART:PreInitialize()
 	self.Children = {}
 	self.ChildrenMap = {}
@@ -62,8 +77,9 @@ function PART:PreInitialize()
 	self.hide_disturbing = false
 	self.active_events = {}
 	self.active_events_ref_count = 0
-	local cvarName = "pac_enable_" .. string.Replace(self.ClassName, " ", "_"):lower()
-	if not GetConVar(cvarName):GetBool() then self:SetWarning("This part class is disabled! Enable it with " .. cvarName .. " 1") end
+
+	local convar = get_enable_convar(self.ClassName)
+	if not convar:GetBool() then self:SetWarning("This part class is disabled! Enable it with " .. convar:GetName() .. " 1") end
 end
 
 function PART:Initialize() end
@@ -81,7 +97,6 @@ end
 
 function PART:GetName()
 	if self.Name == "" then
-
 		-- recursive call guard
 		if self.last_nice_name_frame and self.last_nice_name_frame == pac.FrameNumber then
 			return self.last_nice_name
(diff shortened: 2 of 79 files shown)


lua/pac3/core/client/parts/lock.lua
```diff
@@ -458,7 +461,7 @@ function PART:DecideTarget()
 
 	--sort for the closest
 	for i,ent_candidate in ipairs(ents_candidates) do
-		local test_distance = (ent_candidate:GetPos()):Distance( self:GetWorldPosition())
+		local test_distance = (ent_candidate:GetPos()):Distance(self:GetWorldPosition())
```
thegrb93, line 464: This Distance can be optimized

thegrb93:
I feel like for loops that don't need optimization should use ipairs (which is faster in luajit). Each for loop now being 3 lines is pretty messy and uncomfortable to look at.

TW1STaL1CKY:
I agree it's nicer to look at when it's just ipairs. However when I test the run times between them, `for i = 1, #tbl` is consistently the fastest, even when luajit is enabled.

Results:
```
pairs		0.061870000008639
ipairs		0.023996999989322
for k = 1, #tbl	0.0083710000012616
```

The code I'm using to test the run times:
```lua
local tbl = {"a", "b", "c", "a", "b", "c", "a", "b"}

jit.on() -- Make sure its on

local runTimes = 100

local time = 0
for x = 1, runTimes do
	local start = SysTime()
	
	for i = 1, 1000 do
		for k, v in pairs(tbl) do end
	end
	
	local e = SysTime()
	
	time = time + (e - start)
end

print("pairs", (time / runTimes) * 1000)

time = 0
for x = 1, runTimes do
	local start = SysTime()
	
	for i = 1, 1000 do
		for k, v in ipairs(tbl) do end
	end
	
	local e = SysTime()
	
	time = time + (e - start)
end

print("ipairs", (time / runTimes) * 1000)

time = 0
for x = 1, runTimes do
	local start = SysTime()
	
	for i = 1, 1000 do
		for k = 1, #tbl do
			local v = tbl[k]
		end
	end
	
	local e = SysTime()
	
	time = time + (e - start)
end

print("for k = 1, #tbl", (time / runTimes) * 1000)
```

lua/pac3/core/client/parts/lock.lua
TW1STaL1CKY: True, I missed this file
Thanks 😄

thegrb93:
That makes me really sad for glua. Still, compromising clean code in spots that don't need the performance is a bit of a headache. I'm fine with the changes either way.

CapsAdmin:
I don't wanna dictate too much here as I'm not very active in the project at the moment, but how does this affect startup time, framerate of rendering a complex outfit, and so on?

In other words, if it doesn't improve the performance users care about then all we're really doing is making the code more complex.

Astralcircle:
> I don't wanna dictate too much here as I'm not very active in the project at the moment, but how does this affect startup time, framerate of rendering a complex outfit, and so on?
> 
> In other words, if it doesn't improve the performance users care about then all we're really doing is making the code more complex.

Of most significant, these changes can reduce the number of calls to player.GetAll() and ents.GetAll() using the existing table, for i=1,#tbl can go through the table faster.
Given that in some cases ents.GetAll() and player.Iterator() are called in loops, this can greatly increase performance in them

thegrb93:
That's stating the obvious. He just wants to know if the fps gain is worth the messy code.

pingu7867:
Having to resolve the conflicts in that many files changed will be an absolute pain for me who hoards heaps and heaps of changes before publishing them, but I'll manage.

As for the style, yeah it adds some lines and it's a different way but I'll accept it if it makes pac a little less bulky in the crucial areas that need their inevitable loops. Especially Think and OnDraw ones, that's good to optimize. Even if ultimately there's always a bigger source of cost out there. How much of pac's runtime cost are we fighting over? If it turns out to be less than a percent, fighting over a tiny fraction is useless in my opinion. Just don't make the code unreadable. Which these changes don't.

If we're gonna talk about performance, I'm more interested in numbers and quantifying the impact.

> Of most significant, these changes can reduce the number of calls to player.GetAll() and ents.GetAll(), for i=1,#tbl can go through the table faster. Given that in some cases ents.GetAll() and player.Iterator() are called in loops, this can greatly increase performance in them

Neither the word "can" nor "greatly increase performance" don't mean anything by themselves. For any theoretical "can", you have to show the real truth. I can trust your intent but it's not enough to establish. For any "greatly increase", you must specify how much. What's the run time in milliseconds (or microseconds) before, and what's the runtime after? What's the FPS before and after? If you have numbers, I'm sure they'll speak for themselves.

Astralcircle:
> That's stating the obvious. He just wants to know if the fps gain is worth the messy code. Imo only the hot loops should be optimized, but it doesn't really matter.

Perhaps I didn’t express myself clearly enough. What I meant was that it could improve performance in certain situations. For example, if used on a larger server with a bunch of complex PAC3 outfits, the difference would definitely be noticeable. However, if you were to scrutinize it under a magnifying glass in single-player, there would hardly be any noticeable change.

About complicating the code, I don’t think that anything can complicate it except for k = 1, #tbl, because the rest of the changes simply consist of using a different function

lua/pac3/core/client/base_part.lua
```diff
@@ -53,6 +56,18 @@ function PART:IsValid()
...
+local function get_enable_convar(classname)
+	local convar = pac_enable_convars[classname]
+
+	if not convar then
+		convar = GetConVar("pac_enable_" .. string.Replace(classname, " ", "_"):lower())
+		pac_enable_convars[classname] = convar
+	end
+
+	return convar
+end
```
thegrb93, line 69: This can be the __index metamethod of pac_enable_convars. Then replace calls to this with table indexes.

lua/pac3/editor/client/parts.lua
```diff
@@ -2673,7 +2716,7 @@ function pace.UltraCleanup(obj)
 			if IsSafe(part) or Important(part) then
 				return
 			elseif IsMildlyRisky(part) then
-				if table.Count(part:GetChildren()) == 0 then
+				if #part:GetChildren() == 0 then
```
thegrb93, line 2719: Are we sure these table.Count replacements will always work? vgui can be pretty inconsistent with the type of table this returns iirc.

thegrb93: Maybe use table.Empty instead

thegrb93:
I finished reviewing. Just a couple more suggestions. I think a big potential performance improvement is minimizing __index calls on entities.

lua/pac3/editor/client/parts.lua
CapsAdmin: In theory part:GetChildren() should always return a numerically indexed table. I'm not sure what's up with this code, but I guess in theory is just in theory..

Also I believe [1] is faster than getting the tables length, which part:HasChildren() should do.

wrefgtzweve:
ive seen people complain that player.Iterator ents.Iterator can rarely return NULL entities on clientside, im not exactly how it works and how to fix it but its enough of a issue that i can recommend not using it when possible clientside

lua/pac3/core/client/base_part.lua
thegrb93: Speed doesn't matter for this function. it's just cleaner to use __index. And yeah jit is kinda random about what it blacklists or compiles

lua/autorun/netstream.lua
thegrb93, line 1: Keep this file consistent between addons. It doesn't need optimization.

lua/pac3/editor/client/saved_parts.lua
```diff
@@ -118,7 +118,7 @@ function pace.Backup(data, name)
 		end
 	end
 
-	if #data > 0 then
+	if data[1] then
```
thegrb93, line 121: Should be `if data[1]~=nil then` in case the table contains falses.

lua/pac3/editor/client/saved_parts.lua
```diff
@@ -53,7 +53,7 @@ function pace.SaveParts(name, prompt_name, override_part, overrideAsUsual)
 		override_part = nil
 	end
 
-	if #data == 0 then
+	if not data[1] then
```
thegrb93, line 56: Should be `if data[1]==nil then` in case the table contains falses.

lua/pac3/editor/client/saved_parts.lua
```diff
@@ -63,7 +63,7 @@ function pace.SaveParts(name, prompt_name, override_part, overrideAsUsual)
 
 	data = pac.CallHook("pace.SaveParts", data) or data
 
-	if not override_part and #file_Find("pac3/sessions/*", "DATA") > 0 and not name:find("/") then
+	if not override_part and file_Find("pac3/sessions/*", "DATA")[1] and not name:find("/") then
```
thegrb93, line 66: Add `~= nil`

lua/pac3/core/server/effects.lua
```diff
@@ -47,18 +47,18 @@ net.Receive("pac_request_precache", function(len, pl)
...
 		end
 	else
 		plqueue = {name}
 		queue[pl] = plqueue
 
 		local function processQueue()
-			if #plqueue == 0 then
-				queue[pl] = nil
-			else
+			if plqueue[1] then
```
thegrb93, line 57: Add `~= nil`

lua/pac3/core/client/parts/animation.lua
```diff
@@ -8,7 +8,7 @@ AnimStack = {
 		push = function(self, part)
 			local stack = self.stack
 
-			if #stack == 0 then
+			if not stack[1] then
```
thegrb93, line 11: `stack[1] == nil`

lua/autorun/netstream.lua
```diff
@@ -310,20 +309,18 @@ function net.WriteStream(data, callback, dontcompress)
 		data = util.Compress(data) or ""
 	end
 
-	if #data == 0 then
+	if not data[1] then
```
thegrb93, line 312: `data[1] == nil`

lua/pac3/core/shared/util.lua
```diff
@@ -419,53 +451,57 @@ function pac.DownloadMDL(url, callback, onfail, ply)
...
+									table_insert(files, {file_name = material_name, buffer = dummy, crc = util.CRC(dummy), file_path = material_name})
 								end
 
-								table.insert(found_materials, {name = material_name, offset = material_name_pos})
+								table_insert(found_materials, {name = material_name, offset = material_name_pos})
 								f:seek(material_end)
 							end
 
-							if ply == pac.LocalPlayer and #found_materials == 0 then
+							if is_localplayer and not found_materials[1] then
```
thegrb93, line 478: `found_materials[1] == nil`

lua/pac3/editor/client/saved_parts.lua
TW1STaL1CKY: Good spot, I've applied your suggestion for all these 😄

pingu7867:
I've had some updates I've been meaning to push for a while and had decided to wait. But it's been a month, which is a far cry from what the "within a few days" initially stated. So I'm doing some of my updates and you can deal with the git conflicts later.

TW1STaL1CKY:
> I've had some updates I've been meaning to push for a while and had decided to wait. But it's been a month, which is a far cry from what the "within a few days" initially stated. So I'm doing some of my updates and you can deal with the git conflicts later.

That's fine go ahead. I may exclude my changes to do with combat related things for the time being anyway.

And sorry, there were a lot of things pointed out that I could try improving, so I kept this back to keep working on it.

lua/pac3/core/client/base_part.lua
```diff
@@ -272,7 +276,7 @@ do -- scene graph
 		-- this will handle cases like if a part is removed and added again
 		for _, key in pairs(self.PartReferenceKeys) do
 			if self[key] and self[key].UniqueID == part.UniqueID then
-				self["Set" .. key](self, part)
+				self[prefix_set .. key](self, part)
```
thegrb93, line 279: Making this a variable is pointless. Getting rid of the concatenation would be a good performance improvement.

thegrb93: Also I don't like making these constant strings into variables since it fragments and complicates the code.

thegrb93:
This PR is nearly unreviewable with the amount of changes. You're going to just have to test it on a server probably.

thegrb93:
Conflicts need to be resolved too.

Niterux:
is this already be done or worked on?

pingu7867:
> is this already be done or worked on?

You're rather optimistic. Forgotten is the word you're looking for. Abandoned, even, perhaps. Surely that is the reality you have to recognize when faced with a situation of 20 months of inactivity. But anyway, that's not my PR. What should I think of it when the author of an update doesn't continue to push for it? Really a question I'm often asking myself about our contributors.

PR #1369, opened 2024-06-29
fix outfit filters not working

Techbot121:
the outfit filters were not working unless you opened the settings menu once, which was weird. 
Also it would not ignore them if you requested outfits from the server.

I made a pr because I'm actually not sure about wear_filter.lua, I found it weird how the config file for it was stored in it's own folder separated from the rest.

### DIFF
--- a/lua/pac3/editor/client/wear_filter.lua
+++ b/lua/pac3/editor/client/wear_filter.lua
@@ -3,8 +3,9 @@ local L = pace.LanguageString
 
 local path = "pac3_config/"
 
-do -- old path migration
-	if file.IsDir("pac/config", "DATA") and cookie.GetString("pac3_config_migration_dismissed", "0") ~= "0" then
+-- old path migration
+pac.AddHook("PrePACEditorOpen", "wear_filter_config_migration", function()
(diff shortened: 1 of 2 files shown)


lua/pac3/editor/client/wear_filter.lua
```diff
@@ -1,22 +1,51 @@
...
 local function read_config(id)
-	local tbl = util.JSONToTable(file.Read("pac/config/"..id..".json", "DATA") or "{}") or {}
+	local tbl = util.JSONToTable(file.Read(path .. id .. ".json", "DATA") or "{}") or {}
 	cache[id] = tbl
 	return tbl
 end
 
 local function get_config_value(id, key)
-	return cache[id] and cache[id][key]
+	return cache[id] and cache[id][key] or read_config(id)[key]
```
thegrb93, line 48: Can you set __index metamethod of cache to read_config instead of this?

Techbot121:
I did some testing and it seems like the ignoring part does seem to work without the changes I did to wear.lua, so I'll be reverting that.

thegrb93 (approved):

Issue #1341, opened 2024-03-25
PAC doesnt follow lag compensation in TTT


Earu:
To reproduce this issue, have a player wear an outfit with model parts parented to any bone of a player. Hit the player while they are moving, notice how the part are desync'd.

homonovus:
related source links
https://github.com/ValveSoftware/source-sdk-2013/blob/0d8dceea4310fde5706b3ce1c70609d72a38efdf/sp/src/game/server/player_lagcompensation.cpp#L410
https://github.com/ValveSoftware/source-sdk-2013/blob/0d8dceea4310fde5706b3ce1c70609d72a38efdf/sp/src/game/server/player_lagcompensation.cpp#L611

CapsAdmin:
related garrysmod issue
https://github.com/Facepunch/garrysmod-issues/issues/884

wrefgtzweve:
If i understand the issue correctly the issue is when the player gets lagcompensated.
Did a test with this, seems like only `GetBonePosition` is affected while `GetBoneMatrix` + `GetTranslation` doesnt

https://github.com/CapsAdmin/pac3/assets/69946827/ace5e1d2-533a-4810-8638-2d61124c217f

The test code:
```lua
hook.Add( "HUDPaint", "BoneExample", function()
    for _, ply in ipairs( player.GetAll() ) do
        if not ply:Alive() then continue end
        if ply == LocalPlayer() then continue end

        local bone = ply:LookupBone( "ValveBiped.Bip01_Head1" )
        if not bone then continue end

        local matrix = ply:GetBoneMatrix( bone )
        if not matrix then continue end

        local pos = matrix:GetTranslation()
        local toScreen = pos:ToScreen()

        draw.SimpleText( "GetBoneMatrix", "DermaDefault", toScreen.x, toScreen.y, color_white, TEXT_ALIGN_CENTER, TEXT_ALIGN_CENTER )

        local pos = ply:GetBonePosition( bone )
        local toScreen = pos:ToScreen()

        draw.SimpleText( "GetBonePosition", "DermaDefault", toScreen.x, toScreen.y, color_white, TEXT_ALIGN_CENTER, TEXT_ALIGN_CENTER )
    end
end )
```

TW1STaL1CKY:
Yeah, that is the problem. It looks like PAC does mainly use Entity:GetBoneMatrix, though there are a couple of spots still using Entity:GetBonePosition it seems. Maybe those few instances are the cause?

[...]

PR #1391, opened 2025-02-15
Fixed clip planes lagging behind when parented to a bone on a model part

TW1STaL1CKY:
Resolves last reported problem on issue #1341

### DIFF
--- a/lua/pac3/core/client/parts/bone.lua
+++ b/lua/pac3/core/client/parts/bone.lua
@@ -290,12 +290,12 @@ function PART:GetBonePosition()
 	if not self.bone_index then return ent:GetPos(), ent:GetAngles() end
 
 	local m = ent:GetBoneMatrix(self.bone_index)
-	if not m then return ent:GetPos(), ent:GetAngles() end
 
-	local pos = m:GetTranslation()
-	local ang = m:GetAngles()
-
-	return pos, ang
+	if m then
+		return m:GetTranslation(), m:GetAngles()
+	else
+		return ent:GetPos(), ent:GetAngles()
+	end
 end
 
 function PART:GetBoneMatrix()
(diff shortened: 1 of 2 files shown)


thegrb93:
I really hope this doesn't invoke setupBones for every clip

CapsAdmin:
Just a heads up, I've tried this approach before, and while it seemed to work at first it wasn't exactly a drop in replacement for GetBonePosition. I'm not saying your solution doesn't work. Maybe this solution works now given how things have changed.

TW1STaL1CKY:
> I really hope this doesn't invoke setupBones for every clip

If you mean for the scenario where there's multiple clip parts on one model, you're right and I've now prevented that from happening, since it only needs to be called once.

If you mean for every clip part in general, yes (except for the above scenario now) unfortunately, however the way I see it from briefly benchmarking it, it's faster to just call pac.SetupBones, rather than checking whether the problem is occurring first (requires pac.GetModelBones) then calling pac.SetupBones.

thegrb93:
I still don't like calling SetupBones from a part. It should be done via whatever hook is drawing the parts. Quick hacks like this are what made pac3 so unmaintainable in the first place. If that's not feasible and this does solve the issue then maybe I'd be ok with merging, but bleh.

thegrb93:
I guess the problem is bone parts manipulate the bone and then the clip doesn't have the updated info from the manipulation, which would make my suggestion not feasible. In that case, I'd guess manipulations need to be ordered first in the render stages, then SetupBones, then rendering, but with how pac3 was designed that probably isn't possible. Anyway, you can merge probably.

thegrb93 (approved):

thegrb93:
That said- the bug will still occur if a clip is rendered, and then another bone is manipulated and another clip is rendered.

Issue #1397, opened 2025-04-13
Recieving too many ReadStream requests!


orcineus:
I've been receiving this issue. Using PAC3 on my server, gets around 25-35 people at peak times. 

![Image](https://github.com/user-attachments/assets/02a538f6-98c4-4764-a710-02eb05edd497)

It immediately prompts:

![Image](https://github.com/user-attachments/assets/318ef0fc-3f28-4267-92c3-cff4a14d9a20)

Some users report that they immediately get disconnected despite using pac_onuse 1, which mitigates the issue but not entirely.
Below is another error from another client, after checking the 'Problems' tab in the main menu of Garry's Mod.

```[PAC3] lua/pac3/libraries/string_stream.lua:656: StringStream:readType: Couldn't read type nil
stack traceback:
            [C]: in function 'error'
            lua/pac3/libraries/string_stream.lua:656: in function 'readType'
            lua/pac3/libraries/string_stream.lua:595: in function 'readTable'
            lua/pac3/editor/client/wear.lua:225: in function <lua/pac3/editor/client/wear.lua:218>
            [C]: in function 'xpcall'
            lua/autorun/netstream.lua:355: in function 'ReadStream'
            lua/pac3/editor/client/wear.lua:218: in function 'func'
            lua/includes/extensions/net.lua:37: in function <lua/includes/extensions/net.lua:22>
    1. ReadStream - lua/autorun/netstream.lua:356
        2. func - lua/pac3/editor/client/wear.lua:218
            3. unknown - lua/includes/extensions/net.lua:37

PR #1398, opened 2025-04-14
Improve net error handling

thegrb93:
Fixes: #1397


thegrb93:
Apparently there's some diverged commits

thegrb93:
Need to network the owner of the invalid data so their info is displayed in the console, since it seems to be someone somehow maliciously spamming.

thegrb93:
Tested and working

PR #1401, opened 2025-04-16
Merge develop to master

thegrb93:
About time for an update


thegrb93:
Tested it locally and didn't find any issues

pingu7867:
I was still reluctant (there are still some things I want to sort out), but I guess it's alright since merging into master will give more people the opportunity to / force them to encounter whatever problems might've been outstanding and report them so we can fix it.

Issue #1410, opened 2025-05-19
Loading PAC3 with E.Y.E: Divine Cybermancy mounted causes a CTD


ChrisCj8:
title

idk if this is too specific to consider fixing but i wanted to have it brought up at least

crashlog was taken with all other games and addons (both workshop and legacy) disabled, downloads folder cleared out and all other games dismounted

i've tried renaming the file with the weird space in its name that's mentioned in the crashlog and deleting it and revalidating the game but that didn't fix it

[caef632f-f6d5-4a2d-ae25-3de0319d9936.txt](https://github.com/user-attachments/files/20282440/caef632f-f6d5-4a2d-ae25-3de0319d9936.txt)

[caef632f-f6d5-4a2d-ae25-3de0319d9936.dmp](https://github.com/user-attachments/files/20282433/caef632f-f6d5-4a2d-ae25-3de0319d9936.dmp)

log and dump are from the 64bit branch, but the crash also happens in the normal branch

thegrb93:
Does it crash outfitter too? If so, it might just be something the game can't hot-mount.

ChrisCj8:
oh, i should've probably clarified, the crash happens on the loading screen

E.Y.E: Divine Cybermancy is a source engine game on steam that I've mounted through mount.cfg

thegrb93:
Oh I see, I finally got to look at the txt crash file. It seems some particle systems in that game are attempted to be loaded by pac3 and crashing.

https://github.com/CapsAdmin/pac3/blob/master/lua/pac3/core/server/effects.lua#L14-L22

thegrb93:
I'm not sure how to find the crashing one, but I'd imagine we have to blacklist it to prevent it from being loaded.

Maybe you could replace 'game.AddParticles' with 'print' and tell me what appears in your console?

ChrisCj8:
tried doing that and it still crashed, with the crashlog redirecting me to pac3/core/client/effect.lua this time, so i tried replacing game.AddParticles with print there as well, and that prevents the crash, but i'm not sure how i would figure out what the offending particle is from what it's printed to the console

i also did some experimenting myself after looking at the code you linked and tried to add ALL of EYEs particles to the blacklists (by copying their names over from a file called "particles_manifest.txt" in eyes particle folder) in pac/core/server/effects.lua and pac/core/shared/init.lua and changing the functions i replaced with print back to game.AddParticles but that didn't achieve much

ChrisCj8:
[...]

PR #1413, opened 2025-05-20
Blacklist crashing effect

thegrb93:
Fixes: #1410


thegrb93:
Ah, I see, one is for blacklisted files, and the other is for blacklisted particle effects within those files.

PR #1426, opened 2025-10-15
Fix: Image files served from Google drive do not work

### LINKED ISSUE #1425: Image files served from Google drive do not work - See PR
### **This issue has a PR to fix it:** #1426 

---

At some point in late September 2025 (in my experience, idk when this first started worldwide), Google changed their behavior when serving `https://drive.google.com/uc?export=download&id=aaaaa` style links.

### Issue 1

Interactive loading (i.e. via the pac editor) of remote images is affected.

PAC's `urltex` module uses an HTML document with a simple `` tag to display these images.
Previously, this worked with urls like `https://drive.google.com/uc?export=download&id=aaaaa`. Now, it is broken, because Google refuses to serve the file in the required manner expected by this scenario.

#### Fix

One solution is to use a different url schema: `https://drive.google.com/thumbnail?id=aaaaa&sz=s1024"` for the `src` attribute of the `` tag. Under this schema, when the file is a common image format (like a png, jpg, or gif), Google successfully serves the image in a manner that successfully displays in the `` element.

The `sz` url var specifices a maximum resolution for the served image. It is unclear how exactly Google achieves this, i.e. how their re-encoding process affects the source image. However, they appear to respect the file's instrinsic qualities.

In my very minimal testing with this, I've found that:
- Source PNGs are served as PNGs
- Source JPGs are served as JPGs
- Source GIFs are served as GIFs, animation is preserved
- And in all these cases, the re-encoded image Google serves appears to be faithful to the source image (e.g. color space, apparent quality, gif framerate, etc).
- Uncommon image formats in a Google drive, like VTFs, will _fail_ to serve under the `thumbnail` schema (they are a 404 instead). However, this seems to be a non-problem, since pac needs to download VTFs properly anyways (instead of using the hideous DHTML-with-`` method), which means the `uc?export=download` schema is both appropriate and necessary for VTF and other types like that, and is already expected as such for those cases afaik.
  And this segues into the next problem...

### Issue 2

Non-interactive loading (i.e. on outfit load) of remote images is affected.

PAC's remote resource loading routines fail when the HTTP request serving remote resource has multiple MIME types in the `Content-Type` response header. Google, notably, now includes multiple MIME types when serving files through the `uc?export=download` schema. For example, a PNG file served will have `Content-Type` set to `image/png,application/binary`. In its current state, PAC parses that mime type string into the nonsense file extension `.png,application/binary`, which fails to pass the file extension renaming whitelist and thus blocks `Material()` from working on the downloading file, due to the default `.dat` extension the on-disk file will have instead.

#### Fix

One solution is to appropriately handle multiple MIME types in the `Content-Type` response header. This could be a complex scenario, but for the purposes of pac, it seems sufficient to take the first mime type which passes pac's small and specific whitelist of expected file types. This is the approach taken in the PR. Ultimately, the only thing that matters for images is the presence of a `.jpg` or `.png` extension, regardless of whether the data is actually jpg, png, or something entirely different.

-----

### A Solution

Please refer to the PR which fixes both of these issues.

I hope this issue can be resolved soon for the main workshop release. This issues affects **everyone** who uses Google drive to serve their pac assets.

TiberiumFusion:
Fixes #1425. Refer to explanation of the issue there.

The solution here is straightforward and passes my hasty set of test cases. Further testing is welcome.

### DIFF
--- a/lua/pac3/libraries/resource.lua
+++ b/lua/pac3/libraries/resource.lua
@@ -139,6 +139,7 @@ local function download(from, to, callback, on_fail, on_header, check_etag, etag
 
 	local allowed = {
 		[".txt"] = true,
+		[".jpeg"] = true,
 		[".jpg"] = true,
 		[".png"] = true,
 		[".vtf"] = true,
@@ -150,7 +151,20 @@ local function download(from, to, callback, on_fail, on_header, check_etag, etag
 		function(body, len, header)
 			do
 				if need_extension then
-					local ext = header["Content-Type"] and (header["Content-Type"]:match(".-/(.-);") or header["Content-Type"]:match(".-/(.+)")) or "dat"
+					local ext = "dat"
+					if header["Content-Type"] then
+						for _, mime in ipairs( string.Split(header["Content-Type"], ',') ) do -- some common services, including google, serve multi-item mime strings like "image/png,application/binary"
+							subtype = mime:match(".-/(.-);") or mime:match(".-/(.+)")
+							if subtype then
+								subtype = subtype:lower()
+								if allowed['.' .. subtype] then
+									ext = subtype
+									break -- take the first subtype (if any) found in our whitelist
+								end
+							end
+						end
+					end
+					
 					if ext == "jpeg" then ext = "jpg" end
 
 					if body:StartWith("VTF") then


pingu7867:
I tested it, it works I think.

PR #1436, opened 2025-11-14
Gave physical projectiles radius limit a server convar

lionstrykes10911:
Sinply adds a server variable for the limit of how far away a physical projectile can be spawned, where 0 disables this limit completly.

(sorry if there is any weirdness, I had to do this on the phone.)


lionstrykes10911:
yeah, I didn't have access to the pc rn since my brother was on while I was doing this, I'll make sure to keep it in mind next time.

pingu7867:
Also always push into develop first. Unless it's a high-priority / critical bug fix where we want to bypass the difference between dev and master (i.e. if there are still concerns about stability for certain new features but we still need to fix another thing immediately), then we might do that

I'll see if I can find a better way of syncing dev to master than another PR in the github desktop app. it is cherry-picking!

lionstrykes10911:
noted.

PR #1446, opened 2026-03-21
Damage Zones need nil checks

ByakuyaKuchikie2:
added some nil checks to a few functions. pac3 would spam errors in console from 'Tried to use a NULL entity!' this should fix that issue
image_2026-03-21_140121467


Bonyoze:
Code in here that should not be in PAC3 was merged in, this PR should be reopened or reverted.

PR #1447, opened 2026-03-23
A few more nil checks

ByakuyaKuchikie2:
missed some obvious ones on the first pr. needs to be double checked for any parts with hold entities within them (i.e self.ent)


Bonyoze:
You are including commits that shouldn't be in this PR.

PR #1448, opened 2026-03-26
Revert #1446

Bonyoze:
Merged PR used a branch with leftover changes that should not have been included. It was also merged to `master`. There are better ways to resolve the NULL entity errors so this should really be entirely reverted.


Bonyoze:
It's also merged to `develop`. It should be reverted there too.

PR #1451, opened 2026-04-16
Simple ranger filtering

legokidlogan:
Adds four new boolean options to the ranger event, allowing the trace to ignore the pac owner player, the local player, all players, and/or all npcs. Also makes the npc check respect nextbots.
image

Without this, the only way to filter out your own player entity is by having the event parented to yourself, which prevents you from controlling its position or angle.

### DIFF
--- a/lua/pac3/core/client/parts/event.lua
+++ b/lua/pac3/core/client/parts/event.lua
@@ -1182,7 +1182,12 @@ PART.OldEvents = {
 		arguments = {{distance = "number"}, {compare = "number"}, {npcs_and_players_only = "boolean"}, {ignore_you = "boolean"}, {ignore_viewer = "boolean"}, {ignore_players = "boolean"}, {ignore_npcs = "boolean"}},
 		userdata = {
 			{default = 15, editor_panel = "ranger", ranger_property = "distance"},
-			{default = 5, editor_panel = "ranger", ranger_property = "compare"}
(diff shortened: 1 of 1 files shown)


CapsAdmin:
I wonder if this could be genarlized to a list of things to skip. But I forgot if pac3 already has this concept.

Just a thought, I will not be making the call to merge or not.

pingu7867:
Some general guidelines:

-Prefer merging into dev over master

-Prefer adding a new event instead of overloading an existing event with large numbers of additional arguments. Unless we plan the modularity ahead of time, it ends up a mess especially if we start confusing "only" and "include" types of wording and if options compete. There were granular things we wanted to have so is_touching is now a 4-member series that does slightly different things with each one. It wasn't all pre-planned but we ended up with "good enough" cases that I didn't get pestered with many more requests for a new setting or a new event...

-Prefer absolute backward compatibility unless justified: including nextbots as npcs for the existing npcs_and_players_only option is a change that will happen to existing outfits without the player's input. Does that matter? I don't know. I think it's unlikely we'll get complaints if people seem to treat nextbots as npcs anyway, but keep that in mind. It's not technically 100% backward compatible.

I'm not obsessive over testing things but I trust that your changes work as intended. Approved. I'll wait for your confirmation before merging this. Sometimes people have more changes before they want their PR merged. I'll handle the rest with the dev branch.

PR #1463, opened 2026-07-24
Tools for mirroring parts

Advers:
Provides two tools for quickly mirroring parts across the X, Y, and Z planes. (I tried getting it to support arbitrary planes, but scaling did not work properly.)


pingu7867:
https://github.com/user-attachments/assets/ee3fc914-00b1-4ec1-9e02-11d5e4b4e081

I'll make some additions

PR #1464, opened 2026-07-26
lazy game.AddParticle experiment

Python1320:
game.AddParticle loads all textures into memory from TF2 which is around 700MB VRAM wasted.
This is AI coded experiment to see how well lazy loading would function.

I have really only checked that some sort of particle spawns ingame when I add a effect particle part.
Additional fixes to code required before hopes of merging:  
 - fix lint errors
 - add more error checking to sqlite (full disk will cause lots of error spam)
 - pac code conventions?
 - what if cl.db is completely broken as it is for many users
 - error handling for unknown particle effect parsing
 - maybe try improving particle selector so you can see 
 - Check if serverside particle precaching is required anymore? It seems not as this worked: https://wiki.facepunch.com/gmod/Global.ParticleEffect#example

### DIFF
--- a/lua/pac3/core/client/init.lua
+++ b/lua/pac3/core/client/init.lua
@@ -48,6 +48,10 @@ pac.animations = include("pac3/libraries/animations.lua")
 
 include("pac3/core/shared/init.lua")
 
+pac.pcfprovider = include("pac3/libraries/pcfprovider.lua")
+
+pac.pcfprovider.Initialize()
+
 pac.urltex = include("pac3/libraries/urltex.lua")
 
 include("parts.lua")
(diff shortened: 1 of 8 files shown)


Python1320:
It seems we can maybe get rid of serverside precaching with [CreateParticleSystemNoEntity](https://wiki.facepunch.com/gmod/Global.CreateParticleSystemNoEntity) ( https://github.com/Facepunch/garrysmod-requests/issues/3308#issuecomment-5086274606 )

Python1320:
Vibe-rewrote with [CreateParticleSystemNoEntity](https://wiki.facepunch.com/gmod/Global.CreateParticleSystemNoEntity) with minimal testing (no more serverside components). Will review later, sorry.

Techbot121:
oops didn't mean to do that I reverted that, so this is technically not merged

PR #1467, opened 2026-08-05
Feature: Enhanced Proxies

Wunder-Wulfe:
* Fixed blacklist to use frontier patterns instead of the previous pattern matching structure, which would erroneously detect statements like `owner_velocity_forward()` due to `_for` satisfying the previous `[%s%p]for` match. Now matches against `%f[%a](for)%f[%A]`

* Enables additional compilation mode, allowing use of local variables, if statements, and return statements
  * Provides alternative ways of computing proxies optimally, by re-using calculation results, or short circuit evaluations / early exits via if statements
  * Prohibits function declarations, loops, and modification of global variables for security/safety reasons. Modifying, assigning to, or creating globals will result in an error

* Improved readability / structure of expressions.lua code
  * Moved repetitive code into functions, and simplified logic
  * Additional logic moved into functions to make iteration and testing easier

<details>
<summary>Working Example</summary>

Proxy:
```lua
local r,g,b = 0,0,0

local walkspeed = sqrt(owner_velocity_forward()^2 + owner_velocity_right()^2)

r = (if_event "Grounded")

if walkspeed > 0.2 then
    g = 1
else
    g = 0
end

return r,g,b
```

GIF:
4425fdc4cd5520765e08a46e90468c45

</details>


Wunder-Wulfe:
If the diff doesnt match develop, I can remove the PR and try again with a cleaner fork of the develop branch

PR #1469, opened 2026-08-12
Update text.lua

lionstrykes10911:
Why wasn't this variable used? it's literally the only time it appears here.


lionstrykes10911 (commented):
Everything seems to work great here, did tests with large text amounts and everything runs just fine.

